PR #11, opened 2015-01-28
Scope fix

michalsanger:
The ```didError``` method was called in wrong scope on XHR error


jklein:
Whoops, thanks for fixing that!

PR #36, opened 2015-01-29
Fixes #33, continue when a repo is empty and just report it

### LINKED ISSUE #33: Houndd crashes on empty repositories
One of my github projects is an empty repository at the moment. The `houndd/main.go` crashes every time it tries to scan it: 

```sh
go run src/hound/cmds/houndd/main.go                                                                                                                                (no jobs) [master]  Week 05 29.01.15 13:53
2015/01/29 13:54:00 Searcher started for gol
panic: exit status 1

goroutine 1 [running]:
main.main()
	/tmp/Hound/src/hound/cmds/houndd/main.go:304 +0x5a6
exit status 2
```

Please integrate a check for empty repositories.

jklein:


jklein:
Hmmm, actually hold off on this @kellegous , I think I found a bug

jklein:
Okay should be all better @kellegous

kellegous:
Oops, I think I just conflicted with this when I merged the vcs stuff. @jklein What do you think about just not hiding any VCS output. `cmd.Stdout = os.Stdout` and `cmd.Stderr = os.Stderr` until we can work on that page that shows the indexing status of each repo? I'm torn on skipping errors on startup because it means the repo will be silently absent from the indexes (sadly, it can currently be silently out of date), but I'm cool with it if you think it's the right way to go before we have an index status page.

jklein:
Well in this PR I tried to make it so that we show all VCS output in the event of an error, and don't show anything if it proceeded as planned. I like that approach to avoid polluting the output. 

I think skipping errors on startup is fine as long as we say that it was skipped. While I'd love to surface this in the UI, I think we're better off for now not having the app crash when one repo is empty or fails to pull or something.

src/hound/cmds/houndd/main.go
```diff
@@ -190,14 +191,13 @@ func makeSearchers(
...
-			return nil, err
+		if err == nil {
+			m[strings.ToLower(name)] = s
 		}
-		m[strings.ToLower(name)] = s
 
 	}
 
-	return m, nil
+	return m, err
```
kellegous: Instead of returning the last error, you should probably return a []error or maybe even a map[string]error that maps Repo to error.

src/hound/git/git.go
```diff
@@ -42,7 +36,11 @@ func HeadHash(dir string) (string, error) {
 }
 
 func Pull(dir string) (string, error) {
-	if err := pull(dir); err != nil {
+	cmd := exec.Command("git", "pull")
+	cmd.Dir = dir
+	out, err := cmd.CombinedOutput()
```
kellegous: Instead of combining the output, why don't we just give stdout and stderr of the parent process to the child. `cmd.Stdout = os.Stdout` and `cmd.Stderr = os.Stderr`. That way the streams will be intact.

src/hound/cmds/houndd/main.go
jklein: Yeah, although currently we don't actually display this error anywhere. Previously it was just "exit status 1" anyway.

PR #38, opened 2015-01-29
Fix houndd server location output line

rhysr:
When running houndd the output status line doesn't handle --addr with a hostname included.
I've never written any go before, so this might be complete nonsense :)

**Before**
./bin/houndd
All indexes built, running server at http://localhost:6080...

./bin/houndd --addr 127.0.0.1:6800
All indexes built, running server at http://localhost127.0.0.1:6800...

**AFTER**
./bin/houndd
All indexes built, running server at http://localhost:6080...

./bin/houndd --addr 127.0.0.1:6800
All indexes built, running server at http://127.0.0.1:6800...

### DIFF
--- a/src/hound/cmds/houndd/main.go
+++ b/src/hound/cmds/houndd/main.go
@@ -304,7 +304,13 @@ func main() {
 		panic(err)
 	}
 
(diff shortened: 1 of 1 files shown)


jklein:
Thanks!

rhysr:
No problem, it's a cool project!

PR #35, opened 2015-01-29
add ability to specify branches - fixes #9

caglar10ur:

PR #39, opened 2015-01-30
Refactor internal VCS api and add Mercurial support

belak:
This was actually fairly painless. I refactored all repo accesses to use a new `hound/vcs` package (it replaces `hound/git`) which works similar to `database/sql` in that you can register plugins to handle operations.

The only odd thing is that in stead of doing `hg id -i`, I used `hg log -r tip --template '{node}'` which should give a 40 character sha, not the shorter one, as there's a regex used in existing that's somewhat lazy in looking at directory names.

Also, most of the hg plugin was a direct copy from the git plugin with a few renames. It might be worthwhile to clean it up a bit.

The config adds a new vcs option which allows you to specify the vcs of a repository. It defaults to git.

Docs have not been updated to say that mercurial is supported yet, as I'd like to get some feedback on this first.

### DIFF
--- a/src/hound/config/config.go
+++ b/src/hound/config/config.go
@@ -6,11 +6,15 @@ import (
 	"path/filepath"
 )
 
-const defaultMsBetweenPoll = 30000
+const (
+	defaultMsBetweenPoll = 30000
+	defaultVCS           = "git"
+)
 
 type Repo struct {
 	Url            string `json:"url"`
 	MsBetweenPolls int    `json:"ms-between-poll"`
+	VCS            string `json:"vcs"`
 }
 
 type Config struct {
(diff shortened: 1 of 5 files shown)


kellegous:
This looks great. I'm going to play with it a little bit and then merge it.

kellegous:
@belak If I land https://github.com/etsy/Hound/pull/35, do you have the bandwidth to add the branch support for mercurial?

kellegous:
@belak Nevermind, I think I'm going to land this and see if #35 is interested in changing the way branches work in that change. I'm going to play with this and try to get it merged.

kellegous:
Thanks for doing this. I'm going to try this out on some mozilla repos this weekend. I went ahead and merged this and we can update docs and example configs in separate changes.

afriza:
2015/02/03 14:30:04 Searcher started for ABC-Project
    2015/02/03 14:30:04 Failed to clone file://abc/project, see output below
    Cloning into 'ABC-Project'...
    fatal: '/abc/project' does not appear to be a git repository
    fatal: Could not read from remote repository.

    Please make sure you have the correct access rights
    and the repository exists.
    Continuing...
    2015/02/03 14:30:04 Some repos failed to index, see output above
    2015/02/03 14:30:04 running server at http://localhost:6080...

The folder exists and I have the access right since I'm running it as root.. It is just that it is local mercurial repo.. I am using commit d2c1d98deb169f2539668d6992207a9e9d83b8dd

PR #40, opened 2015-01-30
Add basic SVN support.

MoriTanosuke:
I really like the UI of Hound and the simplicity to set up using the Dockerfile. But I need a code search that will also work on SVN repositories so I started hacking on Hound.

To get started, I added SVN support for all URLs that do not end with `".git"`. Not very clever and golang probably has better ways to deal with this. I'm not very familiar with Golang, so it might be pretty ugly. If you have any tips for improvement or something that I really messed up, please contact me.


MoriTanosuke:
This will collide with PR #39

MoriTanosuke:
I added the VCS api from #39 to my code.

src/hound/vcs/svn.go
```diff
@@ -0,0 +1,59 @@
...
+	par, dirName := filepath.Split(dir)
+	log.Printf("Checkout into %s from %s...\n", dir, url)
+	cmd := exec.Command(
+		"svn",
+		"checkout",
+		url,
+		dirName)
+	cmd.Dir = par
+	cmd.Stdout = ioutil.Discard
+	//TODO do not ignore out, but it's always returning exit status 1 - normal for SVN?
```
kellegous: This was definitely not the norm for SVN when I used it a few years ago; I can't imagine them changing that because it would break all kinds of scripting.

MoriTanosuke: @kellegous You are right, exit code of the commands is 0. I'll look for the mistake I made in the code.

MoriTanosuke:
I fixed the issues with the command exit status and added a test file `svn_test.go` to help me out with XML parsing of the response of `svn info`.

Can anyone have a look at my use of the `struct` to parse the XML? Am I doing this right? :question:

pinky07:
hi

src/hound/searcher/searcher.go
```diff
@@ -133,6 +133,7 @@ func New(vcsDir string, repo *config.Repo) (*Searcher, error) {
 	log.Printf("Searcher started for %s", name)
 
 	sha, err := vcs.PullOrClone(repo.VCS, vcsDir, repo.Url)
+
```
jklein: This looks like an empty change - can you just pull this out since this file doesn't have any changes otherwise?

src/hound/vcs/svn_test.go
```diff
@@ -0,0 +1,48 @@
+package vcs
```
jklein: Thanks for the test, although I think we'd rather have a more holistic approach to testing if we are going to add it. Maybe just pull this out for now until we have figured out what we are going to do here.

MoriTanosuke:
@jklein Will do, probably tomorrow. Time is short today. :bus:

jklein:
Sorry for the delay on this @MoriTanosuke. It looks pretty good - can you just fix the conflict and then we'll merge?

src/hound/vcs/svn.go
```diff
@@ -0,0 +1,98 @@
...
+
+	// TODO parse xml output
+	info := Info{}
+	if err := xml.Unmarshal(buf.Bytes(), &info); err != nil {
+		log.Printf("error: %v", err)
+		return "", err
+	}
+
+	hash := strings.TrimSpace(info.Entry.Repository.Uuid)
+	log.Printf("hash: %s", hash)
```
kellegous: Be sure to clean up the debugging stuff.

Also, there is a bit of a bad spot in our code right now in that we assume that hashes are 40 character hashes that consistent of [a-f0-9]. I need to clean it up, but for now it persists. See this:
https://github.com/etsy/Hound/blob/master/src/hound/searcher/searcher.go#L54

PR #50, opened 2015-01-31
Allow borders in .match div grow to the end of line

michalsanger:
The ```matches``` div had purpose, sorry :-)

screen shot 2015-02-01 at 00 22 57


kellegous:
Thanks.

PR #51, opened 2015-02-02
Add an element for user to hide advanced menu

linc01n:
Provide a more visible way to hide the advanced menu.

hound_hide_adv


jklein:
Thanks!

PR #52, opened 2015-02-02
Update README.md

reinhardt1053:
Cloning the project using https

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Hound is an extremely fast source code search engine. The core is based on this 
 
 This is the preferred approach, since the binaries are generally easier to work with, and make will build both the server and the CLI binaries at the same time. 
 
-1. Clone the repo: `git clone git@github.com:etsy/Hound.git`
+1. Clone the repo: `git clone https://github.com/etsy/Hound.git`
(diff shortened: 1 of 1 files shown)


jklein:
Why is this a change that's worth making? If people want to clone it with HTTPS it's easy enough to do from the GitHub UI.

reinhardt1053:
Well, people not familiar with git that will follow your step-by-step instructions will get an error.

jklein:
Because windows doesn't support SSH style URLs? Sorry I'm just trying to figure out when the error occurs.

reinhardt1053:
screen shot 2015-02-02 at 16 03 21

reinhardt1053:
From Github official doc: 
https://help.github.com/articles/which-remote-url-should-i-use/

"Cloning with SSH

These URLs provide access to a Git repository via SSH, which is a secure protocol. To use these URLs, you must have an SSH keypair generated on your computer, and attached to your GitHub account."

jklein:
Ah okay, got it. Can you just change the other instance in the README as well in that case? I'm happy to merge, but this line appears twice now.

reinhardt1053:
Ok, I am gonna do that :+1:

PR #55, opened 2015-02-02
create parent directory of git repo

chulkilee:
When a repo name has a slash, it failed to clone since the parent directory does not exist.

### DIFF
--- a/src/hound/vcs/git.go
+++ b/src/hound/vcs/git.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"io"
 	"log"
+	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


kellegous:
Hm, I think the right solution to slashes in the name is to escape the Name instead of creating an interstitial directory as a side-effect of the slash. I really should have disallowed slashes in repo names in this first implementation as we really didn't design for it.

chulkilee:
It works without any problem - I could search and the result shows correctly in the UI.

However I agree that if that key is "repo name" then it should not allow slash there. Probably better to have a "group" of repos instead of this kind of implicit nesting - then we may add group-level features, such as limit the search by group.

If you think that's the way to go, feel free to close this merge request and create new issue for that.

kellegous:
Names are no longer in the filenames in the dbpath, so we should be all good.

PR #56, opened 2015-02-03
Adding Windows Support - Refactored code to fix a few compilation issues...

sriniprash:
* Incorporated the comments given for the previous PULL Request.
* The changes are compiling & build is working fine in windows.

### DIFF
--- a/src/code.google.com/p/codesearch/index/mmap_bsd.go
+++ b/src/code.google.com/p/codesearch/index/mmap_bsd.go
@@ -12,6 +12,17 @@ import (
   "syscall"
 )
 
+// An mmapData is mmap'ed read-only data from a file.
+type mmapData struct {
+  f *os.File
+  d []byte
+  o []byte
+}
+
+func Munmap(o []byte) error{
+  return syscall.Munmap(o)
+}
+
 func mmapFile(f *os.File) mmapData {
   st, err := f.Stat()
   if err != nil {
--- a/src/code.google.com/p/codesearch/index/mmap_windows.go
+++ b/src/code.google.com/p/codesearch/index/mmap_windows.go
@@ -11,6 +11,29 @@ import (
 	"unsafe"
 )
 
+// An mmapData is mmap'ed read-only data from a file.
+type mmapData struct {
+  f *os.File
+  d []byte
+  o uintptr
+}
+
+func Munmap(o interface{}) error{
+	// This Munmap function is called from read.go like Munmap(o unitptr) and
+	// from write.go like Munmap(o []byte)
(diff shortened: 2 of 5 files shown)


kellegous:
@sriniprash I'm going to take a look a this, I just have to bring my windows VM back up so I can try it out.

src/code.google.com/p/codesearch/index/mmap_windows.go
```diff
@@ -11,6 +11,29 @@ import (
...
+	// For now, only the calls from read.go are handled using the windows equivalent of Munmap
+	// i.e syscall.UnmapViewOfFile(o uintptr)
+	// Calls from write.go get a nil always.
+	// TODO: Find a way to get the mmapData.o field in write.go.
+
+	switch o.(type) {
+		case uintptr:
+			return syscall.UnmapViewOfFile(o.(uintptr))
+		default:
+			return nil
```
kellegous: Do I read this correctly? On Windows, we would not be unmapping the regions associated with the index writers?

kellegous: I think the right answer here is to have the index writers also hold a proper reference to the whole mmapData and not just the slice taken from the mmap region. Maybe it's time to have mmapData be an io.Closer so we don't have to keep up with what field needs to be passed to Munmap.

src/code.google.com/p/codesearch/index/mmap_bsd.go
```diff
@@ -12,6 +12,17 @@ import (
   "syscall"
 )
 
+// An mmapData is mmap'ed read-only data from a file.
```
kellegous: We might be able to move the _bsd and _linux code into a single mmap_unix.go file.

src/code.google.com/p/codesearch/index/mmap_windows.go
sriniprash: Yea. That's right. Currently it was not possible to unmap the regions in windows due to lack of enough info in mmapData to do so.

src/code.google.com/p/codesearch/index/mmap_bsd.go
sriniprash: Yea. This should be possible.

src/code.google.com/p/codesearch/index/mmap_windows.go
sriniprash: The irst approach (holding mmapData reference) looks doable. Could you please elaborate on the second one?

ameerkat:
Does this build completely? I had to modify the the ansi/tty files as well on my machine.

PR #62, opened 2015-02-05
Set correct CORS header in API endpoint.

fxb:
The correct header string is 'Access-Control-Allow-Origin'.

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS


jklein:
Whoops, thanks!

PR #63, opened 2015-02-06
Support bzr repositories + add a "name" field.

massix:
Hello,
we're starting to use this tool for our codebase but we're using bzr as a VCS.

This pull request adds support for bzr repositories (via the json configuration) and adds also an optional field called "name" that allows a different name to be displayed in the HTML page, this is due to the fact that bazaar branches have weird names that might be misleading, for example we have repo1/releases/2015-02 and repo2/releases/2015-02 and Hound's default behavior is to show only the latest part of the url (in this case "2015-02" which is not that helpful).

Feel free to merge if you're interested.

### DIFF
--- a/pub/assets/js/hound.js
+++ b/pub/assets/js/hound.js
@@ -265,6 +265,11 @@ var Model = {
       return repo;
     }
 
+    // Prefer the given name
+    if (info.name) {
+      return info.name;
+    }
+
     var url = info.url,
         ax = url.lastIndexOf('/');
     if (ax  < 0) {
--- a/src/hound/config/config.go
+++ b/src/hound/config/config.go
@@ -13,6 +13,7 @@ const (
 
 type Repo struct {
 	Url            string `json:"url"`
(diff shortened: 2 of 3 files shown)


jklein:
I don't really like the name option, since that's kind of the purpose of the key in the config. I'd prefer if we just had the JS use that by default instead of the last portion of the URL. For example, you can have something like:

```
{
    "dbpath" : "data",
    "repos" : {
        "ThisIsjQuery" : {
            "url" : "https://www.github.com/jQuery/jQuery.git"
        }
    }
}
```
Right now Hound will show "jQuery" in the UI next to search results, but it should probably show "ThisIsjQuery". This way you can customize the name of the repo there and we don't need another element in the config to do so. 

Can you make this change as part of this PR?

massix:
Of course, but I won't be able to do that until monday :smile:

massix:
Closing this PR in favor of a new one.

PR #65, opened 2015-02-10
Support bazaar's repositories + use json's repo's section name as name

massix:
Similar to #63 with the only exception that the name that will be shown in the HTML interface will be the one used in the .json file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ hound.sublime-workspace
 data
 .vagrant
 pub/index.html
+config.json
--- a/pub/assets/js/hound.js
+++ b/pub/assets/js/hound.js
@@ -259,28 +259,6 @@ var Model = {
     });
   },
 
-  NameForRepo: function(repo) {
-    var info = this.repos[repo];
-    if (!info) {
-      return repo;
-    }
-
-    var url = info.url,
-        ax = url.lastIndexOf('/');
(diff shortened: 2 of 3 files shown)


jklein:
LGTM @massix, thanks. It is going to change the UI for pretty much everyone since we are changing the way we display repos, but I think this will be fine. For most people I think the name in the config matches the repo name anyway. 

@kellegous - take a look and merge when ready.

kellegous:
@massix Can you recommend a public bzr repo to try this out on?

massix:
Hi @kellegous , I am sorry but I'm not masochist enough to use bazaar outside of the working environment, therefore I don't know any public repository :disappointed: 

That said, I think that Canonical's Launchpad hosts some bzr repo, but I'm not sure.

belak:
http://en.wikipedia.org/wiki/GNU_Bazaar#Projects_using_Bazaar

lp:beautifulsoup looks promising.

jklein:
@massix sorry for the delay on this, want to fix the conflict and we'll take another look and hopefully merge soon?

kellegous:
@massix See if there is anything else I need to add in #89.

massix:
Best regards,
Massimo Gengarelli
M: +33 (0)6 43 87 57 67

PR #69, opened 2015-02-12
grep-like and emacs mode

joetastic:
Hey guys!  I don't really know go, but this is pretty simple so hopefully its tolerable :smile:

I just threw the `hound.el` at the top level directory, not sure if you want to add a `contrib/` or somethin.

Hope all is well!

### DIFF
--- a/hound.el
+++ b/hound.el
@@ -1,0 +1,22 @@
+(defgroup hound nil
+  "Variables related to hound."
+  :prefix "hound-"
+  :group 'tools)
+
+(defcustom hound-command "hound"
+  "The name of the hound program."
+  :type '(string)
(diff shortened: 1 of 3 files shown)


jklein:
We're not going to merge `hound.el`, because there's actually a full emacs plugin for Hound being written at Etsy that should be open sourced soon in its own repo, and I don't want to fill up this repo with editor specific files. We should probably remove the sublime stuff as well.

jklein:
@joetastic I'm going to go ahead and close this, the "official" emacs plugin is linked from our README. If you want to still submit a PR for the CLI stuff just open a new one with only those changes. Thanks!

PR #72, opened 2015-02-14
Move code.google.com/p/codesearch into hound/codesearch to avoid collisi...

kellegous:
...ons during go get.


kellegous:
This was pointed out in https://github.com/etsy/Hound/issues/71. If we're going to fork codesearch (which we are for now), we have to pull it into our namespace. I'm going ahead and making this consistent with everything else to make it easier to move around when I do the changes to make the go-gettable.

jklein:
:+1: I think this is a positive change. It's going to break some PRs (specifically the compilation on windows one I think), but overall I think it's time to fork this for real, so I'm into it.

kellegous:
Oh good point on the windows thing. Maybe I'll see about merging that first.

PR #76, opened 2015-02-16
Add patterns to config object for configurable URLs

paulwellnerbou:
See https://github.com/etsy/Hound/issues/75

### DIFF
--- a/pub/assets/js/hound.js
+++ b/pub/assets/js/hound.js
@@ -284,12 +284,13 @@ var Model = {
   UrlToRepo: function(repo, path, line) {
     var info = this.repos[repo],
         url = info.url.replace(/\.git$/, ''),
-        anc = line ? '#L' + line : '';
+        anc = line ? info.anchorpattern.replace('${line}', line) : '';
 
     // Hacky solution to fix _some_ of the 404's when using SSH style URLs
     url = url.replace("git@github.com:", 'https://www.github.com/');
 
-    return url + '/blob/master/' + path + anc;
+    // I'm sure there is a nicer React/jsx way to do this:
+    return info.urlpattern.replace('${url}', url).replace('${path}', path).replace('${anc}', anc);
   }
 
 };
--- a/src/hound/config/config.go
+++ b/src/hound/config/config.go
@@ -9,12 +9,16 @@ import (
 const (
 	defaultMsBetweenPoll = 30000
 	defaultVcs           = "git"
+	defaultUrlPattern    = "${url}/blob/master/${path}${anc}"
+	defaultAnchorPattern = "#L${line}"
 )
 
 type Repo struct {
 	Url            string `json:"url"`
 	MsBetweenPolls int    `json:"ms-between-poll"`
 	Vcs            string `json:"vcs"`
+	UrlPattern     string `json:"urlpattern"`
+	AnchorPattern  string `json:"anchorpattern"`
 }
 
 type Config struct {
@@ -48,6 +52,12 @@ func (c *Config) LoadFromFile(filename string) error {
 		}
 		if repo.Vcs == "" {
 			repo.Vcs = defaultVcs
+		}
+		if repo.UrlPattern == "" {
+			repo.UrlPattern = defaultUrlPattern
+		}
+		if repo.AnchorPattern == "" {
+			repo.AnchorPattern = defaultAnchorPattern
 		}
 	}
 


pub/assets/js/hound.js
```diff
@@ -284,12 +284,13 @@ var Model = {
   UrlToRepo: function(repo, path, line) {
     var info = this.repos[repo],
         url = info.url.replace(/\.git$/, ''),
-        anc = line ? '#L' + line : '';
+        anc = line ? info.anchorpattern.replace('${line}', line) : '';
```
jklein: Won't this break if anchorpattern isn't set in the config? We need defaults on the JS side as well as the Go side I think.

pub/assets/js/hound.js
```diff
@@ -284,12 +284,13 @@ var Model = {
...
 
     // Hacky solution to fix _some_ of the 404's when using SSH style URLs
     url = url.replace("git@github.com:", 'https://www.github.com/');
```
jklein: If we are going this route I'd like to get rid of this hack as well, and have a host config option as well.

paulwellnerbou: With this way this should be possible already. The Stash URLs for the browser are different than the VCS URLs as well:

VCS (git remote): https://stash.tld/scm/<STASH-PROJECT>/<REPO-NAME>.git
URL in Browser: https://stash.faz.net/projects/<STASH-PROJECT>/repos/<REPO-NAME>/browse/${path}${anc}

So setting urlpattern to https://www.github.com/etsy/Hound//blob/master/${path}${anc} should work.

pub/assets/js/hound.js
paulwellnerbou: Yes, if the server does not provide anchorpattern or urlpattern this will break (same as url).
So this change breaks downwards compatibility, the new front end will break with an old backend.

paulwellnerbou:
I just noticed that the URL of skipped files is build elsewhere, thats not working.

pub/assets/js/hound.js
paulwellnerbou: Regarding "if we are going this way": The clean way would be to indicate a repository layout and define the repository layout elsewhere. So the JSON config would look like:

    {
        "dbpath" : "data",
        "repos" : {
            "SomeGitRepo" : {
                "url" : "https://www.github.com/YourOrganization/RepoOne.git",
                "layout" : "github"
            },
            "SomeStashRepo" : {
                "url" : "https://www.stashdomain.tld/scm/YourOrganization/RepoTwo.git",
                "layout" : "stash",
                "baseUrl" : "https://www.stashdomain.tld",
                "project" : "MyProject",
                "repoName" : "RepoTwo"
            }
        },
        "repositoryLayouts" : {
            "git" : {
                "anchorpattern" : "#L${line}",
                "urlpattern" : "${url}/blob/master/${path}${anc}"
            },
            "stash" : {
                "urlpattern" : "${baseUrl}/projects/${project}/repos/${repoName}/${path}${anc}",
                "anchorpattern" : "#${line}"
            }
        }
    }


But this leads to a configuration complexity, because starting with stash, there are at least two properties more to define for each repository. It will get even more complicated with personal repositories in stash, where the URL pattern is:

    https://stashdomain.tld/users/${MyUserName}/repos/${MyRepo}/browse/${path}${anc}

While the scm URL is:

    https://stashdomain.tld/scm/~${MyUserName}/${MyRepo}.git

So I'd prefer the url pattern configuration per repository.

platan:
Stash repository URL pattern is:
`http://<stashdomain>/scm/<project_key>/<repository_slug>.git`
And URL pattern for browsing is: 
`http://<stashdomain>/projects/<project_key>/repos/<repository_slug>/browse`
I think that we do not have to define extra properties, because we can convert one URL into another based on layout type. Example:
`http://localhost:7990/scm/proj/repo1.git` -> `http://localhost:7990/projects/proj/repos/repo1/browse`

Personally, I would like to define default repository layout, because I have repositories of only one type.

kellegous:
@paulwellnerbou I want to merge this. I'll look at fixing the skipped files. I'd also like to change the config slightly. I think it would be good if the url-pattern is in a single directive, maybe:

```
"url-pattern" : {
  "base-url" : "${url}/blob/master/${path}${anc}",
  "anchor" : "#${line}"
} 
```

paulwellnerbou:
Sounds good, i checked out the branch url-pattern in the main repository and it works fine. I didn't test the excluded files though. Although the confi you gave as example does not work, as the $ are not replaced, it should look like this for a personal stash repositories, for example:

```
"GroovyJMX" : {
		"url" : "https://stash.domain.tld/scm/~wellnerbou/myproject.git",
		"url-pattern" : {
			"base-url": "https://stash.domain.tld/users/wellnerbou/repos/myproject/browse/{path}{anchor}",
			"anchor" : "#{line}"
		}
	}
```

Issue #19, opened 2015-01-28
Make links in search results work for SSH style repo URLs


jklein:
Currently Hound only supports HTTP style URLs for repos, and it would be nice to support SSH as well.

kgodwin:
https://github.com/etsy/Hound/blob/b0f9b34552cf28357b20f9dfeeac9f532a8c8375/src/hound/git/git.go#L52

{
    "dbpath": "data",
    "repos": {
        "repo": {
            "url": "git@gitlab-install:namespace\/repo.git"
        }
}

I'm pretty sure this does work as I've set it up for an internal gitlab install over SSH. It basically will take any uri that Git would I believe.

jklein:
Yeah it clones, but then all of the links from search results are broken I believe.

kgodwin:
Ah, I misunderstood what you were asking about. There was another user with a similar question and I thought you were asking a similar one.

brucestephens:
If you want to get something working quickly (i.e., hack the code for your site) the thing to change
seems to be `UrlToRepo` in `pub/assets/js/hound.js`. (I'm assuming you've got something like gitweb, cgit, etc., set up to point at.)

kgodwin:
Given Hound already clones things into data/[repo-name] and it operates as a webapp, wouldn't it just make more sense to just self-host the not http:// repos so it just displays the local copy the search index is built on? [e.g. file:///, ssh]

brucestephens:
Just having Hound serve source seems like a good option (and likely would be a useful default when
the URL doesn't look like a github one). More generally it's surely useful to be able to link to
some other service (one that has syntax highlighting, version control or cross referencing).

kgodwin:
Maybe the path forward then is:

If [contains github.com] -> Link as per current
if [UrlToRepo: contains alternative-link-scheme] -> Link as per listed link scheme
else -> Self-serve [with syntax highlighting and links to the rest of the project]

?

brucestephens:
I guess so. Maybe there are some other special cases (like github) that are worth doing (gitlab, for example), but maybe not. What you suggest is probably a good compromise: DTRT most of the time, and simple to understand how to change it.

kgodwin:
http://[website-url]/[repo-name]/blob/master/[filename]#L[line-number]

Maybe the UrlToRepo should just support tokens then?

brucestephens:
Dunno. Might be best to do something simple: so just cover cases 1 and 2 and forget about the
more annoying case?

[...]

PR #79, opened 2015-02-26
fix 404s for Github Enterprise repos (see #19)

colbyr:
This is a slightly more flexible solution for #19 than
d2c1d98deb169f2539668d6992207a9e9d83b8dd, which worked for private
github.com repos but not for repos on a different domain.


jklein:
Thanks Colby :+1: :palm_tree: :ship:

colbyr:
np!

PR #80, opened 2015-02-28
Added basic SVN driver

justnom:
Added SVN driver to available drivers - some of the function names don't quite match with the command being run.
Credentials are loaded from `SVN_USERNAME` and `SVN_PASSWORD` environment variables.

### DIFF
--- a/src/hound/vcs/svn.go
+++ b/src/hound/vcs/svn.go
@@ -1,0 +1,79 @@
+package vcs
+
+import (
+	"bytes"
+	"io"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+func init() {
+	RegisterVCS("svn", &SVNDriver{})
+}
+
+type SVNDriver struct{}
+
+func (g *SVNDriver) HeadHash(dir string) (string, error) {
+	cmd := exec.Command(
+		"svnversion")
+	cmd.Dir = dir
+	r, err := cmd.StdoutPipe()
+	if err != nil {
+		return "", err
+	}
+	defer r.Close()
+
+	if err := cmd.Start(); err != nil {
+		return "", err
+	}
+
+	var buf bytes.Buffer
+
+	if _, err := io.Copy(&buf, r); err != nil {
+		return "", err
+	}
+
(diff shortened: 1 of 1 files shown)


src/hound/vcs/svn.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+func init() {
+	RegisterVCS("svn", &SVNDriver{})
+}
+
+type SVNDriver struct{}
+
+func (g *SVNDriver) HeadHash(dir string) (string, error) {
+	cmd := exec.Command(
```
kellegous: I'm in process of fixing something right now that will make it possible to use the output of svnversion. Currently, Hound expects versions to be 40 character hashes. I think svnversion returns svn's serial revision number.

src/hound/vcs/svn.go
```diff
@@ -0,0 +1,79 @@
...
+
+	return strings.TrimSpace(buf.String()), cmd.Wait()
+}
+
+func (g *SVNDriver) Pull(dir string) (string, error) {
+	cmd := exec.Command(
+		"svn",
+		"update",
+		"--username",
+		os.Getenv("SVN_USERNAME"),
```
kellegous: I don't feel great about having this depend on environmental variables when everything else depends on config settings. Is there a particular reason to use the environment?

src/hound/vcs/svn.go
justnom: Ah yes I just found the regex that was checking for a 40 character hash. The output of svnversion will give you the revision number and possible trailing characters if the [working copy is modified](http://svnbook.red-bean.com/en/1.7/svn.ref.svnversion.re.html). 

I didn't think using it would be a problem if the WC was never going to be modified.

src/hound/vcs/svn.go
justnom: I only used the environment variables in this driver as the config struct for the repo was not passed into any of the drivers functions. I could change this though?

kellegous:
@justnom I added the username, password to the config. https://github.com/etsy/Hound/blob/master/config-example.json#L19-L24

justnom:
@kellegous awesome! I was just about to integrate you changes from the config branch into this when I see you've already committed it :) Thanks for the help!

PR #81, opened 2015-03-02
Url patterns

kellegous:
Fixes some issues in the url patterns PR.
1. Changes mini-templating to use {name} instead of ${name}.
2. Merges url-pattern into a single json property.
3. Makes excluded files page respect url pattern.
4. Introduces common.js file to share between pages.

### DIFF
--- a/pub/assets/js/common.js
+++ b/pub/assets/js/common.js
@@ -1,28 +1,31 @@
 // TODO(knorton): Use something to bundle this more intelligently and get this
 // out of the global scope.
 
-var ExpandVars = function(template, values) {
-  for (var name in values) {
-    template = template.replace('{' + name + '}', values[name]);
-  }
-  return template;
-};
+var lib = {
 
-var UrlToRepo = function(repo, path, line) {
-    var url = repo.url.replace(/\.git$/, ''),
-        pattern = repo['url-pattern'],
-        anchor = line ? ExpandVars(pattern.anchor, { line : line }) : '';
+    ExpandVars: function(template, values) {
+      for (var name in values) {
+        template = template.replace('{' + name + '}', values[name]);
+      }
+      return template;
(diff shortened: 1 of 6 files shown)


kellegous:
@paulwellnerbou I changed a few things from #76. Objections?

pub/assets/js/hound.js
```diff
@@ -282,17 +282,7 @@ var Model = {
...
-        anc = line ? '#L' + line : '';
-
-    // Hacky solution to fix _some more_ of the 404's when using SSH style URLs
-    var sshParts = /git@(.*):(.*)/i.exec(url);
-    if (sshParts) {
-      url = '//' + sshParts[1] + '/' + sshParts[2];
-    }
-
-    return url + '/blob/master/' + path + anc;
+    return UrlToRepo(this.repos[repo], path, line);
```
jklein: This is a little confusing, since it's calling a function in the global scope with the same name as this one.

pub/assets/js/common.js
```diff
@@ -0,0 +1,28 @@
...
+        pattern = repo['url-pattern'],
+        anchor = line ? ExpandVars(pattern.anchor, { line : line }) : '';
+
+    // Hacky solution to fix _some more_ of the 404's when using SSH style URLs
+    var sshParts = /git@(.*):(.*)/i.exec(url);
+    if (sshParts) {
+      url = '//' + sshParts[1] + '/' + sshParts[2];
+    }
+
+    // I'm sure there is a nicer React/jsx way to do this:
```
jklein: Yeah I agree...I'll do some reading on this tomorrow and see if I can find something better. Doesn't have to block the merge though.

kellegous:
I went ahead and wrapped the stuff in common.js in its own namespace.

PR #82, opened 2015-03-03
Fix the naming issues in dbpath.

kellegous:
dbpath names are now:
vcs: vcs-{ SHA1 of repo.Url }
idx: idx-{ pseudo-random 64-bits }

The one downside of this change is we will always delete all indexes and rebuild them at startup. This is a thing that can be fixed though by writing a manifest into each of the idx directories.

### DIFF
--- a/src/hound/cmds/houndd/main.go
+++ b/src/hound/cmds/houndd/main.go
@@ -3,7 +3,6 @@ package main
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"flag"
 	"fmt"
 	"hound/api"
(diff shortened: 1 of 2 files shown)


jklein:
I feel like rebuilding all indexes at startup is prohibitively expensive. We need a better deployment strategy to go that route, since it will mean minutes of downtime for each deploy (at least at Etsy, and other orgs with lots of big repos).

kellegous:
I powered through adding manifests. I need to do some more testing, but this represents the basic approach.

kellegous:
This has been merged. https://github.com/etsy/Hound/commit/c7cd2a71dd861bb0faf20943c5e5d57183a861fe

PR #85, opened 2015-03-08
Support linking to bitbucket style URLs

kppullin:
Bitbucket style URLs differ from github style URLs in two ways:

1) Bitbucket SSH repository URLs takes the form of:
`ssh://hg@bitbucket.org/username/Foo` (vs
`git@github.com:username/Foo.git`)
2) Direct links to source file lines require the commit hash/revision
id, following this pattern: `{url}/src/{rev}/{path}{anchor}`.

This patch addresses both these concerns to allow for direct file
linking to bitbucket.

This required a change to bubble up the revision to the UI layer, which
can be then fed in to the `UrlToRepo` function for final formatting via
a custom `url-pattern` (as defined above).

Server side, the `revision` value is bound to each `index` upon
opening. The `index.Search` function returns this value in the
`SearchResults` struct.

### DIFF
--- a/src/hound/index/index.go
+++ b/src/hound/index/index.go
@@ -20,6 +20,7 @@ type Snapshot string
 
 type Index struct {
 	dir string
+	rev string
 	idx *index.Index
 	lck sync.RWMutex
 }
@@ -44,6 +45,7 @@ type SearchResponse struct {
 	FilesWithMatch int
 	FilesOpened    int           `json:"-"`
 	Duration       time.Duration `json:"-"`
+	Revision       string
 }
 
 type FileMatch struct {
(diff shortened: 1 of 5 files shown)


src/hound/index/index.go
```diff
@@ -57,7 +59,7 @@ type ExcludedFile struct {
 }
 
 func (s Snapshot) Open() (*Index, error) {
-	return Open(string(s))
+	return Open(string(s), "")
```
kellegous: This would be problematic if we still used this notion of a Snapshot, but we don't. So I'm going to merge your change and then fix this in a separate change.

kellegous:
Thanks for fixing this.

src/hound/index/index.go
kppullin: Ah yes, thanks for spotting that.  I had originally ripped it out, but put it back in just in case I was missing something... and forgot to follow up on it.

PR #86, opened 2015-03-09
Moves "manifests" into the index package.

kellegous:
- Removes index.Snapshot
- Creates index.IndexRef that is responsible for index metadata.

### DIFF
--- a/src/hound/index/index.go
+++ b/src/hound/index/index.go
@@ -2,6 +2,7 @@ package index
 
 import (
 	"compress/gzip"
+	"encoding/gob"
 	"encoding/json"
 	"fmt"
 	"hound/codesearch/index"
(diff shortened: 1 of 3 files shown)


kellegous:
TODO:
- Tests on index resurrection.
- Tests on index removal.

kellegous:
This just better encapsulates the notion of manifests within the Index, instead of the searcher. I realized this was way cleaner when I landed that last PR. Now I'm going to focus on getting the various VCS additions landed.

jklein:
Looks good @kellegous, and I like that we are starting to have tests. At some point we should probably add TravisCI - it's free for open source projects.

kellegous:
Oh good idea. I should have set that up earlier. We do have a number of tests buried in the index code.

PR #87, opened 2015-03-10
Add the ability to pass through 'vcs-config' into the vcs drivers.

kellegous:
This should give @justnom what he needs for the svn driver and I'm
sure other drivers will need this as well.

### DIFF
--- a/src/hound/config/config.go
+++ b/src/hound/config/config.go
@@ -19,10 +19,11 @@ type UrlPattern struct {
 }
 
 type Repo struct {
-	Url            string      `json:"url"`
-	MsBetweenPolls int         `json:"ms-between-poll"`
-	Vcs            string      `json:"vcs"`
-	UrlPattern     *UrlPattern `json:"url-pattern"`
+	Url            string          `json:"url"`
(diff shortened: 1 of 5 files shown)


src/hound/config/config.go
```diff
@@ -19,10 +19,11 @@ type UrlPattern struct {
...
 
 type Repo struct {
-	Url            string      `json:"url"`
-	MsBetweenPolls int         `json:"ms-between-poll"`
-	Vcs            string      `json:"vcs"`
-	UrlPattern     *UrlPattern `json:"url-pattern"`
+	Url            string          `json:"url"`
+	MsBetweenPolls int             `json:"ms-between-poll"`
+	Vcs            string          `json:"vcs"`
+	VcsConfig      json.RawMessage `json:"vcs-config"`
```
jklein: Can we put an example of how to use this in the example config?

jklein:
Just the one comment, otherwise LGTM

src/hound/config/config.go
kellegous: Yes, but I want to land the svn driver first because it's going to be the first one to use the vcs-config.

PR #88, opened 2015-03-10
Svn driver

kellegous:


jklein:
:+1:

src/hound/config/config_test.go
```diff
@@ -0,0 +1,36 @@
+package config
```
kellegous: Heads up @jklein, if we commit an invalid config-example.json the tests will break.

jklein: Sounds good to me.

PR #89, opened 2015-03-11
Integrate bzr vcs changes from #65.

kellegous:


jklein:
I don't know much about bazaar, but this looks pretty reasonable to me.

massix:
:shipit: seems fine, thanks!

PR #90, opened 2015-03-11
A better fix for the json marshalling/unmarshalling of vcs-config.

kellegous:
The issue I ran into earlier was simple, I needed VcsConfig to be
a *json.RawMessage instead of a json.RawMessage. But I did not want
to just fix that because it would have sent the vcs-config as JSON
to the client exposing any credentials that are stored there. I use
a trick here where I create a special type like json.RawMessage that
will preserve bytes in Unmarshal, but just emit {} in marshal.

### DIFF
--- a/src/hound/config/config.go
+++ b/src/hound/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"encoding/json"
+	"errors"
 	"os"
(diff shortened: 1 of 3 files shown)


jklein:
Good stuff.

Issue #7, opened 2015-01-28
Please make it go get-able


psankar:
Please make it go get-able

pierrre:
```
Set your GOPATH: export GOPATH=`pwd`
```
is wrong.

shoenig:
Yeah, requiring Rake(!) for a Go project is insane

kellegous:
This is definitely on the list. The only blocker as been figuring out what to do with the web assets that Hound needs for the UI. My current plan is to support two modes of building:

(1) When installing with `go get` hound will continue to depend on the src directory that is cloned during the build process.

(2) Make build will aim to build a more stand-alone version where the assets are bundled in some form.

The command line tool may already be go-gettable, but I haven't checked in a while and it's certainly not useful without a server.

pierrre:
https://github.com/jteeuwen/go-bindata and commit the generated .go files?

sriniprash:
https://github.com/mjibson/esc interfaces with net/http

caglar10ur:
In the meantime what about providing a Makefile like following? Rake is giving hard time to the distros with older ruby versions (undefined method `absolute_path' for File:Class)

```
all:
	@GOPATH=`pwd` go build -o bin/houndd src/hound/cmds/houndd/main.go
	@GOPATH=`pwd` go build -o bin/hound -ldflags "-X main.defaultHost ${HOST-localhost:6080}" src/hound/cmds/hound/*.go

.PHONY: all
```

kellegous:
I put up a PR to switch to a Makefile until this is all go-gettable https://github.com/etsy/Hound/pull/44

belak:
Do the maintainers have any preferences for this? Something like go-bindata and a Makefile to turn the assets into go files might be the best option.

caglar10ur:
I suggest taking a look at https://github.com/golang/tools/blob/master/godoc/static/makestatic.go as well. This is what godoc.org uses.

matrixik:
Or use https://github.com/GeertJohan/go.rice and it will work with `go get` without need for commiting generated files.

caglar10ur:
@matrixik but with go.rice, one cannot distribute a binary only file as it requires source code to be available. 

To me, godoc is way simpler but of course YMMV. 

* Put your resources into https://github.com/golang/tools/blob/master/godoc/static/makestatic.go, 
* Create a file like 
https://github.com/golang/tools/blob/master/godoc/static/gen.go so that one can use "go generate" 
[...]

PR #94, opened 2015-03-13
Make Hound Go Gettable.

kellegous:
Sorry this is a bit of a code bomb and it's hard to test in a branch. Here's how I would recommend trying it out.

```
git clone -b go-gettable https://github.com/etsy/Hound.git hound/src/github.com/etsy/hound
cd hound
GOPATH=`pwd` go install github.com/etsy/hound/cmds/...
./bin/houndd
```

Also be sure to try out the new development workflow for the UI. It's documented in the README.md. Instead of a --prod flag, we now have a --dev flag to enable edit/refresh behavior. When running in production, the assets are all encoded in `ui/bindata.go`.

I'm sure I've missed a couple of things, so I'm going to continue poking around and adding more documentation in the code.


kellegous:
Probably obvious, but this is intended to fix #7.

kellegous:
Oh one other thing of note. We currently require Go 1.3.3 and I would like to go to Go 1.4 so I can use `go generate` for the code generation ... but Etsy still needs this to run on Go 1.3 for a bit longer.

belak:
I'm curious what the decision-making process was like on go-bindata vs go.rice (and other similar systems) as I'll probably be doing something like this with one of my projects in the future.

On a side note, this looks awesome. Nice work!

kellegous:
@belak That's a great question. I have to say that I'm using bindata right now, but I would categorize my decision as: it seemed like a good place to start. I have a few things I want to test before I really commit to it. I've used it in other contexts, but never in a server context. To be honest, the primary motivation at this point is that bindata supports a bit of a hack to take the bytes directly out of the .rodata section of the binary without doing a memcpy. Since this is all immutable data, that was appealing to me. I think other packages come equipped with http.FileSystem adapters which is convenient, but Hound didn't really need that.

.gitignore
```diff
@@ -1,8 +1,2 @@
-pkg
-bin
-hound.sublime-workspace
-data
 .vagrant
-pub/index.html
-pub/excluded_files.html
-config.json
```
jklein: We probably want to retain this line, right? For people who are hacking on Hound. bin and *.sublime-workspace should also stay I would think. 

We also probably want to ignore ui/bindata.go

jklein:
Looks solid. I tested it out briefly and everything works. With all of the VCS support, go-gettability, and editor plugins this is becoming a pretty legit project :-). Nice work @kellegous

.gitignore
kellegous: I think I'll leave config.json to be safe. I suspect it's going to be in a different directory for most people now. But it's good to ensure that nobody sends one in. I'll also leave the .sublime-workspace file. I've actually moved mine out of the source directory but maybe I'll symlink it back. I'm not sure how other people hack on github projects, I make my Hound directory into a workspace so that the source is in `src/github.com/etsy/hound` ... which means I'm generally operating from the root of that directory.

`ui/bindata.go` definitely cannot be in the .gitignore. When we change the UI, we have to regenerate that file and check it in so that users who are installing with `go get` will have all the web assets.

kellegous: I mention [workspace](https://golang.org/doc/code.html) casually, but I'm talking the "workspace" taken from the Go docs.

jklein: Ah I see, I just figured that since `make clean` deleted `ui/bindata.go` we wouldn't want to commit it, but that makes sense that if people aren't editing Hound and just want to run it they would need it. My bad.

kellegous: Oh great point, `make clean` really shouldn't delete it. That was only useful while I was getting it working.

PR #95, opened 2015-03-15
Fixes #91.

### LINKED ISSUE #91: The advanced filter show handler doesn't seem to respond to keyboard tab events
Tabbing allows you to type into the File Path without seeing what you're typing because the filter doesn't show when the File Path input field gets focus from tab.

kellegous:
When the files element receives focus just expand the advanced panel.


jklein:
:+1:

Issue #105, opened 2015-04-07
houndd: don't start polling until all indexes are built


youknowjack:
I was trying to bootstrap hound with 100s of projects, and the polling running simultaneous with index creation meant the server wouldn't start for hours. I have a patch that doesn't start polling until right before the http server is started, after all the searchers are created. That dramatically improves startup time when working with many projects.

PR #106, opened 2015-04-07
Issue #105 - don't start polling VCS until all indexes are built

youknowjack:
See https://github.com/etsy/Hound/issues/105

### DIFF
--- a/searcher/searcher.go
+++ b/searcher/searcher.go
@@ -22,6 +22,8 @@ type Searcher struct {
 	idx  *index.Index
 	lck  sync.RWMutex
 	Repo *config.Repo
+	vcsDir string
+	name string
 }
 
 /**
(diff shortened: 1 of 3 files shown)


searcher/searcher.go
```diff
@@ -211,6 +211,12 @@ func MakeAll(cfg *config.Config) (map[string]*Searcher, map[string]error, error)
 	return searchers, errs, nil
 }
 
+func PollAll(cfg *config.Config, searchers map[string]*Searcher) {
```
kellegous: This looks pretty good, but PollAll shouldn't be an exported function as there is currently no use for a Searcher that is not actively polling. Instead of exposing the notion of an inactive Searcher it would be better to have MakeAll construct all the searchers (without the background poller, like you have it) and then start the polling on each of them at the end of that function.

Something like this maybe:

```
for _, searcher := range searchers {
  go searcher.poll()
}

return searchers, errs, nil
```

youknowjack: Sounds better, agreed. Made the suggested change. How do you feel about me adding vcsDir and name to the searcher struct?

kellegous:
This seems reasonable. There is no need to regenerate bindata.go for this change, though, since there are no changes to the UI. I'll merge this tonight and revert he bindata. Thanks for fixing this, this is pretty big for the growing number of people who have lots of repos.

kellegous:
@youknowjack Sorry for the slow response. I checked out the PR and I see now there were some tricky parts to separating the creation from the polling. As I looked at it, I realized it was probably better to keep the code somewhat isolated as it is now, but introduce a channel to hold the polling go-routines at bay until all the searchers were setup. So I created an alternative PR at #107. See if this works for you.

PR #107, opened 2015-04-14
An alternate approach to #105.

kellegous:


youknowjack:
Cool. Seems much more like the Go Way. Thanks.

PR #109, opened 2015-04-15
Index dot files by default but expose option to config.

kellegous:


config-example.json
```diff
@@ -6,7 +6,8 @@
         },
         "AnotherGitRepo" : {
             "url" : "https://www.github.com/YourOrganization/RepoOne.git",
-            "ms-between-poll": 10000
+            "ms-between-poll": 10000,
+            "excluded-dot-files": true
```
jklein: Should this be "exclude-dot-files" to match the Go and be a little more intuitive?

config/config.go
```diff
@@ -25,6 +25,7 @@ type Repo struct {
 	Vcs              string         `json:"vcs"`
 	VcsConfigMessage *SecretMessage `json:"vcs-config"`
 	UrlPattern       *UrlPattern    `json:"url-pattern"`
+	ExcludeDotFiles  bool           `json:"exclude-dot-files"`
```
jklein: Yeah this makes me think the above is just a typo.

config-example.json
kellegous: Thanks, that's definitely a typo.

jklein:
LGTM

PR #110, opened 2015-04-15
Fixes a bug where an empty list of excludes is sent as null.

kellegous:
A quirk of the way that the json is serialized for the excluded files: if the list of excluded files happens to be empty, the JSON that gets serialized is simply `null` instead of an empty list. This causes an exception in the UI. This just ensures that `[]` is returned if there are no excludes, instead of `null`.

### DIFF
--- a/index/index.go
+++ b/index/index.go
@@ -16,8 +16,9 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


jklein:
LGTM

PR #114, opened 2015-04-18
Allow pushing updates instead of polling - fixes #111

### LINKED ISSUE #111: Allow pushing updates instead of polling
For a large number of repos, it would be nice to be able to turn off polling and hit an API endpoint to update the indexes instead.  This would reduce unnecessary traffic and keep results more recent.

I have an implementation of this that I can submit a PR for if this seems reasonable.

jnackman:


kellegous:
This is a really nice looking change, thanks. I'm going to try this out tomorrow and get it landed.

kellegous:
I'm preparing to land this but I wanted to give you some heads up on some changes that I've made:

1. I changed the configs from `push-enabled` and `poll-enabled` to 'enable-push-updates' and `enable-poll-updates`.
2. I have restricted `/api/v1/updates` to only `POST` since it doesn't match the semantics for `GET`. Let me know if this is problematic.
3. I was a little worried about the fact that repeated calls to `/api/v1/update` would continue to enqueue goroutines so I have introduced an atomic to coalesce update requests if there is already one pending. The practical effect is that if and update is currently underway and you request 100 updates, only one will trigger after the current update finishes ... there will also be a single goroutine blocked on updateChan in that period (instead of 100).
4. I've restructured the code a bit in places just to keep things consistent with existing code and to reuse some of your newly introduced code in other places.

jnackman:
Thanks for the update.  All of those changes sound great to me!

Issue #113, opened 2015-04-17
Error: too many open files after indexing large number of repos


ibam:
We tried using Hound on our company's codebase, but there seem to be an upper limit to the number of repos or the size of codebase that Hound can handle. In our case, we are indexing ~250 repos containing a combined total of millions of LOCs, and Hound (or the underlying Go) emitted the following error when we tried to execute a search:

> ERROR:open /Users/iarief/workspace/hackathon-april2015/hound/data/idx-7e26ac6530283f3e/raw/content.txt: too many open files

The total size of the data folder is about 5.5 GB when compressed. Is this beyond the known practical upper limit of Hound?

kellegous:
With that many repos, you have have to tweak your file descriptor limits. I'm really curious what your current limit is?

From the same login you run Hound, what do these commands report?
`ulimit -Sn`
`ulimit -Hn`

The searchers are currently setup to be pretty aggressive in terms of parallelism ... we can probably put a practical limit on that to avoid opening so many files simultaneously when working on a large repo list.

kellegous:
I setup a test this weekend with 1k repositories and what I'm finding is that I can execute searches just fine without exhausting the file descriptors limit (1024) so long as no indexing is happening. When the searchers are trying to update simultaneously (especially if they are running git), I hit issues. The right answer here, I think, is to actually throttle indexing to ensure that no more than N searchers are reindexing at a time.

PR #116, opened 2015-04-24
[DEVTOOLS-91] Adding connection limiter

mmerrell:
Adding entries to .gitignore to aid in Mac development

[DEVTOOLS-91] Adding connection limiter
[DEVTOOLS-91] Updating README and config-example.json

### DIFF
--- a/searcher/searcher.go
+++ b/searcher/searcher.go
@@ -199,7 +199,7 @@ func init() {
 // NOTE: The keys in the searcher map will be normalized to lower case, but not such
 // transformation will be done on the error map to make it easier to match those errors
 // back to the original repo name.
-func MakeAll(cfg *config.Config) (map[string]*Searcher, map[string]error, error) {
+func MakeAll(cfg *config.Config, connectionLimiter) (map[string]*Searcher, map[string]error, error) {
 	errs := map[string]error{}
 	searchers := map[string]*Searcher{}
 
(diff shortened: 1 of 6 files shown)


mmerrell:
This PR should address issue-115

kellegous:
Also #113. Thanks so much for fixing this, I skimmed the change and it looks good. Token bucket in a channel was exactly what I planned to do. I'm a little behind in integrating PR's but I hope to get caught up very soon.

searcher/searcher.go
```diff
@@ -199,7 +199,7 @@ func init() {
 // NOTE: The keys in the searcher map will be normalized to lower case, but not such
 // transformation will be done on the error map to make it easier to match those errors
 // back to the original repo name.
-func MakeAll(cfg *config.Config) (map[string]*Searcher, map[string]error, error) {
+func MakeAll(cfg *config.Config, connectionLimiter) (map[string]*Searcher, map[string]error, error) {
```
kellegous: Finally got around to merge this tonight, but I'm finding it doesn't build.

kellegous:
@mmerrell  Can you take a look at #118 and make sure it still solves the problem you are having. I changed a couple of things:

 1. The concurrency limit is about more than just the outgoing connections. We need a limiter on indexers in general. So I have renamed the config to `max-concurrent-indexers`.
 2. I move the creation of the token bucket out of main and put it into the searcher. I also created a type to wrap the channel so we can have `Acquire` & `Release` semantics. I also moved a good chunk of the reindexing loop into its own function to make this a bit cleaner.

PR #117, opened 2015-04-25
Small changes to push update coalescing, config defaults

jnackman:
Implements comments on 23e8dbf6b1457ca78e6dad2f1ee2b0b8a9794065.


kellegous:
That's awesome. I've been looking for a way to detect would-block cases in Go channels and I had no idea that select supported that.

kellegous:
Thanks again.

PR #125, opened 2015-05-31
Hide Advanced search after query submit

pvamshi:
We have around 50+ projects and its taking whole page to display them. So, we need to scroll down a lot to view the results .

### DIFF
--- a/ui/assets/js/hound.js
+++ b/ui/assets/js/hound.js
@@ -335,6 +335,7 @@ var SearchBar = React.createClass({
       break;
     case 13:
       this.submitQuery();
+      this.hideAdvanced();
       break;
     }
   },
@@ -453,7 +454,7 @@ var SearchBar = React.createClass({
               onKeyDown={this.queryGotKeydown}
               onFocus={this.queryGotFocus}/>
           <div className="button-add-on">
-            <button id="dodat" onClick={this.submitQuery}></button>
+            <button id="dodat" onClick={this.submitQuery&&this.hideAdvanced}></button>
           </div>
         </div>
 
--- a/ui/assets/js/hound.js
+++ b/ui/assets/js/hound.js
@@ -335,7 +335,6 @@ var SearchBar = React.createClass({
       break;
     case 13:
       this.submitQuery();
-      this.hideAdvanced();
       break;
     }
   },
@@ -362,6 +361,7 @@ var SearchBar = React.createClass({
     this.showAdvanced();
   },
   submitQuery: function() {
+    this.hideAdvanced();
     this.props.onSearchRequested(this.getParams());
   },
   getRegExp : function() {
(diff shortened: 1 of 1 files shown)


kellegous:
I'll check this out, but I'm skeptical that this is the right solution. By hiding the advanced search view when you are doing an advanced search, you hide the state of the current search essentially introducing a hidden mode. This means we now need some additional UI to communicate the fact that advanced filters are active. To be honest, the list of repos was never given much thought and shouldn't be allowed to take up so much vertical space. I would prefer to simply limit the amount of vertical space used by the list of repositories.

kppullin:
What might be a reasonable height to restrict the list? We have 250+ repos in our list and have had a couple people ask to hide the box after searching.

Thinking further ahead, after restricting the vertical height I'd anticipate the next issue will be that it's difficult to know how many repos are selected, if any.  Perhaps a little text saying something along the lines of 'X of Y selected', 'All', or...?

kellegous:
In the short term, limiting the size to around 8 would be ok. Finding the repository in the big list is only marginally worse when having to scroll. My long term preference is to just create a different kind of control. It should have type-ahead and by default it should just show the repos that are selected (with controls for dismissing them). Perhaps something like Google+'s mechanism for adding circles and people. I've also been thinking about a way for power users to change advanced options without leaving the query text field ... which would be nice for people who use the advanced options a lot in their daily work.

gwindes:
What about adding some sort of persistent preferences. Either hound instance specific like the repository list or user specific via some new settings page? There are a couple things like ignoreCase and autoHide on the advanced options dropdown that could be thrown into a preference file/page. Thoughts?

kellegous:
@gwindes Is it that you want to change the default for ignoreCase? I'm not against some kind of config for some of this stuff, but I was hoping we could find a solution better than auto-hiding the advanced options. Hiding the active filters in a search is generally a big usability issue (i.e. unexpected poor recall). Can you tell me more about the problem you are fixing here? Maybe we can come up with something.

BTW, I also wanted to mention (since it's not really documented) that ignore-case searches can be significantly slower. We have actually considered making ignore-case the default, but we didn't want to do it until we changed the index to better support these searches.

gwindes:
@kellegous Yes being able to change the default for ignoreCase. For instance some team members have expressed interest in ignoreCase being defaulted to true. I figured a settings/pref the user can set would alleviate this problem. For example, when I am searching something I may not know the exact case and thus ~95% of my searches are with ignore case enabled (even when using a local grep tool). While it may take longer to get my results I've found it better than having to redo my search with ignore case enabled. 

I suspect the initial reason for wanting to auto-hide the advanced options was due to the repo list size taking up the entire page (with 200+ repos). This caused you to scroll to the bottom half of the page select your repos and then scroll back to the top in order to perform a search and close the advanced options. Maybe the max size on the repo list will alleviate that pain point. 

I agree just hiding the advanced options might not be the best option. Maybe adding some additional information like @kppullin mentioned about which repos are currently selected would be a good addition.

PR #127, opened 2015-06-09
Max size for repo list on advanced options

gwindes:
Some users have massive amounts of repos 200+ which forces the advanced option dropdown to take up the entire screen. This will limit it to a size of minimum(16, repoCount) to avoid the list taking the entire screen.


gwindes:
Related to #125 

Somewhat of a compromise.

PR #128, opened 2015-06-11
Add --ignore-externals support to SVN driver

stuartloxton:
We've got a few separate projects all with the same external library pulled in via svn externals. This adds the option to checkout and update without the externals so you don't get duplicate results.

### DIFF
--- a/vcs/svn.go
+++ b/vcs/svn.go
@@ -15,8 +15,9 @@ func init() {
 }
 
 type SVNDriver struct {
-	Username string `json:"username"`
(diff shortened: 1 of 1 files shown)


kellegous:
Does ignore-externals even need to be a configuration option? Seems like we should just ignore-externals always.

stuartloxton:
I guess not, or at least defaulting the other way might be better. I was expecting it to not include externals when I was setting it up but don't know about other people using the SVN driver. Let me know if you want me to update the branch.

PR #129, opened 2015-06-11
Implemented user preferences for setting ignore case and auto hide advanced on search.

gwindes:


gwindes:
Related to pull request #125

kellegous:
I really don't think it makes sense to add the complexity of user preferences for a single setting. For one, you can no longer depend on links to searches to be stable. A link for `?q=Println` will now not work for some people. I also still don't think that hide advanced makes sense as an option. The core issue is that it was not scaling properly, which was a bug. It's also a thing that we can continue to make better with UI work. I definitely don't want to enable hidden modes in the UI for our own use.

Why don't we start with a server config to change the default for ignore case?

ui/assets/js/cookies.js
```diff
@@ -0,0 +1,63 @@
...
+|*|
+|*|  A complete cookies reader/writer framework with full unicode support.
+|*|
+|*|  Revision #1 - September 4, 2014
+|*|
+|*|  https://developer.mozilla.org/en-US/docs/Web/API/document.cookie
+|*|  https://developer.mozilla.org/User:fusionchess
+|*|
+|*|  This framework is released under the GNU Public License, version 3 or later.
+|*|  http://www.gnu.org/licenses/gpl-3.0-standalone.html
```
kellegous: I'm not keen on adding GPL licensed libraries. If we can avoid it, I'd rather not take on the complexity of having more than one license.

ui/assets/js/cookies.js
```diff
@@ -0,0 +1,63 @@
...
+          sExpires = "; expires=" + vEnd.toUTCString();
+          break;
+      }
+    }
+    document.cookie = encodeURIComponent(sKey) + "=" + encodeURIComponent(sValue) + sExpires + (sDomain ? "; domain=" + sDomain : "") + (sPath ? "; path=" + sPath : "") + (bSecure ? "; secure" : "");
+    return true;
+  },
+  removeItem: function (sKey, sPath, sDomain) {
+    if (!this.hasItem(sKey)) { return false; }
+    document.cookie = encodeURIComponent(sKey) + "=; expires=Thu, 01 Jan 1970 00:00:00 GMT" + (sDomain ? "; domain=" + sDomain : "") + (sPath ? "; path=" + sPath : "");
```
kellegous: The server wouldn't need to know about these preferences, so there is no need to use cookies. localStorage would be more appropriate.

gwindes: Good point. I don't do a ton of work in JS so cookie was my first thought. I'll look into local storage if we decide this is useful.

gwindes:
I initially thought about a server config but decided on the individual user approach due to our case where we have 100+ devs in the company (maybe 25 using our hound instance) and they may not agree on the default. This way each user can customize it to their liking.

Also this preference feature isn't necessarily only for ignore case. Other team members mentioned having a pref for file filters such as excluding .xml files etc 

Regarding the link sharing where it's just `?q=Println`, on a query hound creates all params, while still possible to only share the query param probably unlikely. However, I agree that should be fixed and could be an easy fix (if no url param for ignoreCase override to 'unchecked'). Unless there is a way to create a query without all the params in the url?

@kellegous does your team typically share hound search links? I haven't come across it here (but that doesn't mean it doesn't happen) and doesn't mean this tweak should break that functionality.

PR #133, opened 2015-06-16
Use golang code standards for import blocks.

kellegous:
https://github.com/golang/go/wiki/CodeReviewComments#imports


jklein:
Simple enough, ship it!

PR #136, opened 2015-06-21
Use onbuild

sstarcher:
If we use an ONBUILD instruction this container can be automatically build via docker.io
example https://registry.hub.docker.com/u/sstarcher/hound/

Once that is happening instead of cloning this repo and adding a config.json locally you could do the following.

Dockerfile
```
From sstarcher/hound
```

config.json
```
{ ....}
```

Whatever your config is
Those 2 files would be enough to run hound.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM golang
 
 COPY . /go/src/github.com/etsy/hound
-COPY config.json /hound/
+ONBUILD COPY config.json /hound/
 RUN go-wrapper install github.com/etsy/hound/cmds/houndd
 
 EXPOSE 6080
--- a/README.md
+++ b/README.md
@@ -13,7 +13,10 @@ Hound is an extremely fast source code search engine. The core is based on this 
 
 1. Create a [config.json](config-example.json) in a directory with your list of repositories.
 
-2. Run ```docker run -it --rm -p 0.0.0.0:6080:6080 --name houndd -v $(pwd):/hound etsy/hound```
+2. Run 
+    ```
+    docker run -it --rm -p 0.0.0.0:6080:6080 --name houndd -v $(pwd):/hound etsy/hound
+    ```
 
 You should be able to navigate to [http://localhost:6080/](http://localhost:6080/) as usual.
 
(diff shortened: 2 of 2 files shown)


kellegous:
Be patient with me. Sadly, I know very little about Docker. How do the commands in the README need to be adjusted to use this Dockerfile?

kellegous:
So every time I run the image, it is unable to find `/hound/config.json`. When does ONBUILD run?

sstarcher:
On build runs when you reference the image that has the onbuild instruction.
1. So you would push etsy/hound to docker.io
2. Someone would reference etsy/hound at the top of their Dockerfile
3. When this person builds their docker image based off of etsy/hound the ONBUILD instructions will now run.

The difference in the workflow would be you would not clone this repo.  You would create your own Dockerfile and reference etsy/hound along with dropping your config.json into the directory with your Dockerfile.

sstarcher:
@kellegous  let me know if you have any other questions or want any more concrete examples.

kellegous:
@sstarcher Thanks.

The biggest issue is that the instructions in the REAME need to work with the Dockerfile. Not being much of a Docker user, I'm not sure what the norm is for distributing an image. I have been sitting on this PR until I figure out what other projects are doing here.

sstarcher:
I updated the documentation on this branch to reflect one option.

To test this out you would have to replace etsy/hound in docker run with sstarcher/hound as I have the image on Docker IO.

On my branch the current quick start reflects the following.  This using docker's volume capability to mount the config.json file into the container.
1. Docker is required
2. Add the config file and run docker

The option I talk about in this pull request at the top is similar, but instead of mounting a docker volume.  We build a new container and use the etsy/hound aka sstarcher/hound docker container as the base.  When the container builds it uses the ONBUILD instruction to add the local config.json into the container image itself.

Both of these options have the added benefit that we don't have to have go installed locally.

kellegous:
Sorry for the delay, I'm just getting around to setting up the etsy/hound repository on DockerHub. The process is incredibly flaky but I think I finally have an automated build working. I'm about to merge this.

Issue #134, opened 2015-06-19
Support indexing local directories (without supported VCS)


KindDragon:
I want use Hound as https://github.com/google/codesearch/ for my local directory with code (it subdirectory of my TFS repository)

jklein:
We should probably just detect if the `file://`protocol is being used and bypass the VCS code entirely, just indexing the repo.

aidansteele:
@jklein I opened up a PR that hopefully addresses this issue: #138.

jamime:
Same issue here. I want to index my already checked out SVN repo.

rkachowski:
This would be great - I have a huge mercurial repo with the largefiles extension already checked out that i'd like to be able to index directly

lukehsiao:
Indexing local directories would be a great feature. Is there any way to workaround this while the PR is being updated?

xdsv:
+1 on local directories

PR #138, opened 2015-06-28
Added "none" type VCS for indexing local directories

aidansteele:
This addresses #134. Please go easy on me, this is the first Go I've ever written and it may well be the worst you've ever seen. :smiley: 

I used the git VCS driver as inspiration and basically cut it done to nothing. I figured this approach was better than special-casing `file://` URLs as it a) has one less special case subverting the driver architecture and b) still allows for the scenario wherein a user wants to specify a `file://` URL pointing to a git repo.

I am happy to make any changes as you see fit.

### DIFF
--- a/vcs/none.go
+++ b/vcs/none.go
@@ -1,0 +1,46 @@
+package vcs
+
+import (
+	"log"
+	"os/exec"
+	"path/filepath"
+)
+
+func init() {
+	Register(newNoneVcs, "none")
+}
+
+type NoneVcsDriver struct{}
+
+func newNoneVcs(b []byte) (Driver, error) {
+	return &NoneVcsDriver{}, nil
+}
+
+func (g *NoneVcsDriver) HeadRev(dir string) (string, error) {
+    return "n/a", nil
+}
+
+func (g *NoneVcsDriver) Pull(dir string) (string, error) {
+	return g.HeadRev(dir)
+}
+
+func (g *NoneVcsDriver) Clone(dir, url string) (string, error) {
+	par, rep := filepath.Split(dir)
+	cmd := exec.Command(
+		"cp",
+		"-r",
+		url[7:],
+		rep)
+	cmd.Dir = par
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		log.Printf("Failed to clone %s, see output below\n%sContinuing...", url, out)
+		return "", err
+	}
+
+	return g.HeadRev(dir)
+}
+
+func (g *NoneVcsDriver) SpecialFiles() []string {
+	return []string{}
+}


KindDragon:
:+1:

vcs/none.go
```diff
@@ -0,0 +1,46 @@
...
+	par, rep := filepath.Split(dir)
+	cmd := exec.Command(
+		"cp",
+		"-r",
+		url[7:],
+		rep)
+	cmd.Dir = par
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		log.Printf("Failed to clone %s, see output below\n%sContinuing...", url, out)
```
jklein: I would change this log statement since we are doing a copy not a clone

vcs/none.go
```diff
@@ -0,0 +1,46 @@
...
+func (g *NoneVcsDriver) Pull(dir string) (string, error) {
+	return g.HeadRev(dir)
+}
+
+func (g *NoneVcsDriver) Clone(dir, url string) (string, error) {
+	par, rep := filepath.Split(dir)
+	cmd := exec.Command(
+		"cp",
+		"-r",
+		url[7:],
```
jklein: This seems a little odd...hardcoding `7` here strips off the `file://` I assume? I'd prefer to do a string replace I think.

vcs/none.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (g *NoneVcsDriver) Pull(dir string) (string, error) {
+	return g.HeadRev(dir)
+}
+
+func (g *NoneVcsDriver) Clone(dir, url string) (string, error) {
+	par, rep := filepath.Split(dir)
+	cmd := exec.Command(
+		"cp",
```
jklein: As mentioned in the comment this will break on Windows.

kellegous:
Strangely, I was just discussing this idea on Friday and proposed a similar solution. That said, I think there are a few high-level things we should reconsider:

1. We don't actually gain anything by copying the file contents into the vcs directory. I was thinking maybe we could restructure the vcs driver to return the directory that should be used for reindexing. That way this vcs driver could return the direct path to the local directory.

2. By default, it seems like the none-type vcs should not do polling. Since we know the user is likely to be mutating the local directory, we should require the user to request the update with the new push updates mechanism. (Note that there isn't a great way to trigger this right now, but we could fix that).

3. I've been trying to think of the right name for this vcs driver. 'none' doesn't properly convey what is going on. local? raw? Maybe it does, thoughts? @jklein

jklein:
Yeah I agree with the first two comments from @kellegous, and naming wise I sort of like "file" or "local". It doesn't matter a whole lot though since it should be clear from the documentation.

vcs/none.go
aidansteele: A string replace would be much better! I just hadn't looked up how to do it in Go.

vcs/none.go
aidansteele: Good point. I will look through the Go stdlib to see if there's a cross-platform way to do this.

vcs/none.go
aidansteele: Will do

aidansteele:
Thanks to everyone who has responded so quickly with useful feedback. @kellegous I agree with your first and second points entirely. I didn't make those changes myself because a) it's not my place to make large changes without consulting you first and b) I wouldn't have been up to the task anyway. :smile: 

Regarding naming, "none" is a bit poor. "File" sounds good. "Local" might be a bit misleading as you can have a local git repo.

danielcb:
Hi, do you plan on implementing the changes/features mentioned by @kellegous anytime soon? I would do it myself, but I'm not really familiar with Go. Especially 1. would be really nice.

aidansteele:
Hi @danielcb, sorry I didn't respond sooner. I'm not sure if your comments were addressed to me, but if they were: I'm not really up to the task of implementing those changes, which is why I haven't done it. If it was one of the other guys, don't mind me.. :)

aparcar:
Ping?

PR #141, opened 2015-07-04
Adds tools to make local building much more sane.

kellegous:
Lots of people who do not have deep familiarity with the Go tools are trying to make little changes to Hound and having trouble figuring out the various build commands. I've also been using a similar Makefile since we made Hound go-gettable and I've been looking for a good way to share it with others. This makes the local development flow a lot more opinionated, but it should work.


jklein:
Looks good to me!

PR #142, opened 2015-07-07
fix failing build due to case-sensitive path etsy/Hound

douglas-larocca:
The last commit 753b605719f14087672ccb9bd5bef20a83953721 adding a build setup script raises an issue with the case-sensitive path. The diff on the build instructions:

 ```
-git clone https://github.com/etsy/Hound.git hound/src/github.com/etsy/hound
+git clone https://github.com/etsy/Hound.git hound/src/github.com/etsy/Hound
 cd hound
-GOPATH=`pwd` go install github.com/etsy/hound/cmds/...
+src/github.com/etsy/Hound/tools/setup
+make
 ```

There were two choices: stay with the old path (lowercase) and change the Makefile template and build directions, or globally change everything to use the uppercase name. I went for the latter considering it might be better to remove the ambiguity entirely.

All tests pass. Attached image shows a sort of combined test & diff (couldn't resist).

hound


kellegous:
Thanks. Sadly, I think this should go the other way. The hound repo was accidentally created with a capital H. I need to see if I can undo that somehow.

kellegous:
Ha, renaming the repo was trivial. Let me now fix the build instructions.

kellegous:
@douglas-larocca can you try #143 and see if it fixes things for you. I'm on a mac at the moment with a case-insensitive filesystem.

douglas-larocca:
(The throwaway commit ebc7365f410ec9bf9bed70b1171732cf7bf30070 just tests that the case change doesn't affect git remote settings.)

PR #146, opened 2015-07-24
Refactor Hound's web UI to accept non-regex queries

jonlai:
Currently, Hound only allows for regex searches, both in the query and the file path. This is an inconvenience if the user is unfamiliar with regex, or is simply copy-pasting lines with special characters. This pull request offers non-regex options by encoding/decoding the query to/from regex for the API. 

To implement this feature, we broke up the UI into more React components to modularize the logic and restyled the search panel to accommodate the options.


dgryski:
This would be easier to review if the commits were squashed into a smaller number of logical changes.

jklein:
Agreed, if you could rebase/squash into one commit that would be great.

jonlai:
Sorry about that @dgryski and @jklein, we just squashed all our commits into two -- the first one represents a refactor into multiple React components, and the second one represents the additional search functionality we implemented.

jklein:
Thanks @jonlai - it might be a couple days before I get a chance to pull this down and try it out, but in principle the option seems worth having. Any initial thoughts @kellegous?

kellegous:
I do have some feedback, but I'm going to be delayed on this as well as I'm going to be offline for the next couple of days.

PR #148, opened 2015-08-08
Added filename as an anchor parameter for BitBucket line link support.

dharmabruce:
Sample link to a line in a file of a bitbucket repository:
https://bitbucket.org/fboender/bbcloner/src/master/src/bbcloner#bbcloner-32


jklein:
I assume this is tested and works with anchor patterns that have no filename? If so, then it looks good to me. @kellegous what do you think?

dharmabruce:
Yes, I've tested it with anchor patterns that have no filename and paths with single to several levels of directories.

ui/assets/js/common.js
```diff
@@ -13,7 +13,8 @@ var lib = {
     UrlToRepo: function(repo, path, line, rev) {
         var url = repo.url.replace(/\.git$/, ''),
             pattern = repo['url-pattern'],
-            anchor = line ? lib.ExpandVars(pattern.anchor, { line : line }) : '';
+            filename = path.replace(/^.*\//, ''),
```
kellegous: I don't understand the intent of this line. Is this to strip the trailing "/"?

kellegous: Oh I see, this is extracting the basename. Seems like there is a more straightforward approach with `lastIndexOf`. `path.substring(path.lastIndexOf('/')+1)`?

dgryski: http://golang.org/pkg/path/#Base ?

dharmabruce:
I agree that the suggestion clarifies the intention. I will push an update today.

ui/assets/js/common.js
dharmabruce: I agree that the suggestion clarifies the intention. I will push an update today.

davidmankin:
I've tested this and it works for me.  Would be great to see it merged.

kellegous:
Sorry for the delay.

Issue #150, opened 2015-08-17
Support linking to a GitHub wiki page


miketheman:
Since GitHub wikis are repositories as well, it felt only natural to try using hound to index them for search as well.

config example:
```json
        "datadog/dd-agent.wiki": {
            "url": "https://github.com/DataDog/dd-agent.wiki.git"
        },
```

Indexing, searching all works nicely, however the generated link leads to a 404 page, like so:

https://github.com/DataDog/dd-agent.wiki/blob/master/Home.md#L7

I'd propose that: "if the repo is a github-based repo, and there's a `.wiki` in the repo name, handle the URL differently".

miketheman:
Closed via #155

PR #155, opened 2015-09-13
Introduce support for GitHub wiki search

miketheman:
Simplistic approach in supporting GitHub wiki via frontend replacements
of paths.

Requires a non-default `url-pattern/base-url` in configuration file.

Refs #150 

I thought about going one step further and having the default use case detect if the url passed is a wiki or not, but that seemed to go deeper into changing [this code](https://github.com/miketheman/hound/blob/95306e06a49c8459d5ab372ebcc8bd79707a484c/config/config.go#L104-L112) and it seemed like keeping the change simple was a better approach.


miketheman:
Tagging @kellegous, who is apparently the holder of merges here.

kellegous:
This seems reasonable. I hope to clean up this URL munging on the client-side sometime in the future. I'm going to try this out soon and then I'll get it merged.

avleen:
Tried it out, seems to work without breaking anything. Will merge as it seems pretty useful!

kellegous:
Thanks @avleen, I forgot to merge this.

avleen:
No problem @kellegous! Got your back, you've been busy :-)

PR #156, opened 2015-09-21
Support for local VCS allowing to index local path not necessarily be…

mrKrizz:
Support for local VCS allowing to index local path not necessarily being a part of the version-controller repository.

Non-copying version of https://github.com/etsy/hound/pull/138.


jklein:
This looks good to me, thanks. I'll let @kellegous take a look and merge when ready.

kellegous:
@gemignani Sorry, I have been really behind on pull requests. I'll take a closer look at this soon. I'll leave some minor comments now. I think the one thing that is missing in the local support is that local vcs really shouldn't poll by default because it introduces the possibility that the repo will be indexed while it is being mutated. However, that's a thing I can easily help out with.

searcher/searcher.go
```diff
@@ -367,6 +365,13 @@ func newSearcher(
 		SpecialFiles:    wd.SpecialFiles(),
 	}
 
+	var vcsDir string
+	if (wd.IsLocal()) {
```
kellegous: I think this would better separate concerns if the vcs interface had a function kind of like this:

```
WorkDirFrom(dbpath string, repo *config.Repo) string
```

vcs/local.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (g *LocalDriver) Clone(dir, url string) (string, error) {
+	// For local driver Clone() is only called when the directory
+	// pointed by url is not found.
+	err := fmt.Errorf("Location %s not found.", url)
+	fmt.Print(err)
+	return "", err
+}
+
+func (g *LocalDriver) IsLocal() (bool) {
```
kellegous:
```
func (g *LocalDriver) IsLocal() bool {
```
Single value returns do not have ()

vcs/local.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (g *LocalDriver) Pull(dir string) (string, error) {
+	return g.HeadRev(dir)
+}
+
+func (g *LocalDriver) Clone(dir, url string) (string, error) {
+	// For local driver Clone() is only called when the directory
+	// pointed by url is not found.
+	err := fmt.Errorf("Location %s not found.", url)
+	fmt.Print(err)
```
kellegous: Did you mean to leave the Print?

vcs/vcs.go
```diff
@@ -24,6 +24,9 @@ type Driver interface {
 	// Return the revision at the head of the vcs directory.
 	HeadRev(dir string) (string, error)
 
+	// Returns whether provider operates directly on local file sytem url
+	IsLocal() (bool)
```
kellegous: No parens around `bool`

vcs/local.go
mrKrizz: yes, I saw other drivers both return error and print some log.

searcher/searcher.go
mrKrizz: like in in vcs.Driver?
Then each driver (apart from local) will need to implement basically the same code to get the working dir.

vcs/local.go
kellegous: All the other drivers use `log` to report the actual error. `fmt.Print` is not only going to give an inconsistent log line, but it also won't print a \n at the end of the line.

searcher/searcher.go
kellegous: Right. I'm saying that Driver should not have `IsLocal`. Rather than having the `Searcher` have special knowledge of what `IsLocal` means, it would be better to just move the path creation into the driver. That does mean that this new method will have identical implementations for all the existing drivers, but that's also the case with `IsLocal`.

snagytx:
Any plans to finish this support?

mrKrizz:
@snagytx, I'd love to, but very short time recently. :(

PR #159, opened 2015-09-26
Minimal docker image

dnephin:
Hello,

Not sure if you're interested in this or not, but I worked on it for myself, so I thought I'd share.

This change splits the `Dockerfile` into two, one for building the binary (static) and another that uses a small base image for running the binary.  With this setup the deployed runtime image is 34M (the current one with golang base is ~500M I believe.

The Makefile supports a fully dockerized build process.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .build
 .DS_Store
 *.exe
+dist/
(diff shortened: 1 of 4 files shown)


sstarcher:
+1

PR #165, opened 2015-10-18
Update to React 0.14.0

Friss:
Thoughts @jklein @kellegous? 

Also going to do some other UI things but this seemed like a good place to start


Friss:
Also working on updating everything to work with React 0.14.0. Bunch of places need to be updated to the new React API.

ui/assets/excluded_files.tpl.html
```diff
@@ -9,8 +9,11 @@
     <body>
         <div id="root"></div>
 
-        <script src="js/react-{{.ReactVersion}}.min.js"></script>
+        <script src="js/react-with-addons-{{.ReactVersion}}.min.js"></script>
+        <script src="js/react-dom-{{.ReactVersion}}.min.js"></script>
         <script src="js/jquery-{{.jQueryVersion}}.min.js"></script>
+        <script src="js/babel-core.5.8.23.browser.min.js"></script>
+        <script src="js/common.js"></script>
```
kellegous: Previously, in production builds, we were fetching 2 external scripts and then had the whole of the main script inline. We're now fetching 5 isolated scripts. I'd rather bundle some more scripts together than to continue accumulating more HTTP requests. Also, are all of these files needed in production builds? I had assumed that babel, for instance, would only be needed in `--dev` mode.

Friss: The babel core is replacing the JSX inline transpiler. This can be skipped entirely by transpiling the JSX and only sending that. 

We could also probably combine React with addons and React Dom together in one file if necessary.

ui/ui.go
```diff
@@ -73,14 +73,11 @@ func renderForDev(w io.Writer, root string, c *content, cfg *config.Config, r *h
 	}
 
 	var buf bytes.Buffer
-	fmt.Fprintf(
-		&buf,
-		"<script src=\"/js/JSXTransformer-%s.js\"></script>\n",
```
Friss: @kellegous This is what is being replaced by babel-core.

kellegous: Right. JSXTransformer was previously only added in Dev mode because Dev mode was the only time we needed runtime transformation of JSX. In Prod mode, we used the command line `jsx` to bundle raw javascript in bindata.go (this is located in the `Makefile`).

Friss: Heh TIL. I'll move the files around.

Friss:
Updated the file includes to reflect the Dev vs Prod environment.

Issue #151, opened 2015-09-01
Hound's UI has poor accessibility


tgt:
I find the UI of Hound to be extremely difficult to read since its text is very low contrast. I've noted several of the issues I've encountered below. These issues won't just affect the visually impaired, but also those with lower quality screens.

The guidelines laid out in the [Web Content Accessibility Guidelines (WCAG) 2.0](http://www.w3.org/TR/WCAG20/) suggest a contrast ratio of 4.5:1 for AA (3:1 for text of at least 18px) and 7:1 for AAA (4.5:1 for text of at least 18px). I think the default UI should aim for AAA compliance in order to be accessible to as many people as possible.

Examples:
- The text around the search box (e.g. the advanced section, number of results) has a contrast ratio of 2.32:1 (#aaa on #fff), which fails guidelines for text of any size.
- The text of file paths has a contrast ratio of 5.27:1 (#666666 on #f5f5f5), which fails the guideline for text under 18px.
- The text in the Select Repo section of the advanced options fails with a contrast ratio of 5.74:1 (#666 on #fff), which fails the guideline for text under 18px.
- Line numbers also fail with a contrast ratio of 2:1 (#aaa on #eee), which fails guidelines for text of any size.

jklein:
Thanks, this is a good idea. We'd be happy to accept a PR for these changes. Candidly it's not something we are likely to get to in the near term.

salemhilal:
We’ve just merged in a fix for this issue. I’m closing it out, but feel free to re-open it if something’s missing.

PR #166, opened 2015-10-18
Update style add in Dark theme

Friss:
So this tries to address #151 and make hound a little more customizable. 

Default Theme:
screenshot 2015-10-18 02 48 47
screenshot 2015-10-18 02 49 04

New Dark Theme:
screenshot 2015-10-18 02 48 26
screenshot 2015-10-18 02 49 12

This builds off #165 but I could always redo it off master if you don't want to update React...


kellegous:
I'm pretty skeptical that benefits of adding multiple themes justifies the added burden of maintenance. Why do we need a dark theme?

Friss:
I think that having at least the idea of a light vs dark theme is good for this kind of tool giving the end user some more options is a good thing. Plus hound is pretty bright because it is very white. 

As for the burden of maintenance I abstracted out all of the non theme related CSS to one file while the light and dark files contain only `color` and `background-color` declarations for the most part. This should really cut down on the burden of maintain multiple themes.

PR #167, opened 2015-10-18
Rsync

sturmf:
This is a simple rsync vcs provider. It can be also used for indexing of local directories. Only drawback in that case is that it still makes a copy of the local files.

### DIFF
--- a/vcs/rsync.go
+++ b/vcs/rsync.go
@@ -1,0 +1,45 @@
+package vcs
+
+import (
+	"io/ioutil"
+	"os/exec"
+	"path/filepath"
+)
+
+func init() {
+	Register(newRsync, "rsync")
+}
+
+type RsyncDriver struct{}
+
+func newRsync(b []byte) (Driver, error) {
+	return &RsyncDriver{}, nil
+}
+
+func (g *RsyncDriver) HeadRev(dir string) (string, error) {
+	return "n/a", nil
+}
+
+func (g *RsyncDriver) Pull(dir string) (string, error) {
+	return g.HeadRev(dir)
+}
+
+func (g *RsyncDriver) Clone(dir, url string) (string, error) {
+	par, rep := filepath.Split(dir)
+	cmd := exec.Command(
+		"rsync",
+		"-r",
+		url[7:],
+		rep)
+	cmd.Dir = par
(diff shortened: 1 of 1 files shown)


vcs/rsync.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+type RsyncDriver struct{}
+
+func newRsync(b []byte) (Driver, error) {
+	return &RsyncDriver{}, nil
+}
+
+func (g *RsyncDriver) HeadRev(dir string) (string, error) {
+	return "n/a", nil
```
kellegous: The way this is setup, Hound will continue to try to poll this vcs even though it can never update. I think it would be better to return something sensible for the revision so that pulling works as expected. That said, I do think that an rsync vcs should not poll by default. It should require an explicit push update to tell the server to go fetch the new version. The server supports push updates, but there isn't a convenient way to trigger one right now.

vcs/rsync.go
```diff
@@ -0,0 +1,45 @@
...
+
+func (g *RsyncDriver) Clone(dir, url string) (string, error) {
+	par, rep := filepath.Split(dir)
+	cmd := exec.Command(
+		"rsync",
+		"-r",
+		url,
+		rep)
+	cmd.Dir = par
+	cmd.Stdout = ioutil.Discard
```
kellegous: You shouldn't need to discard this. By default, the subprocesses should not inherit the parent's stdout and stderr.

vcs/rsync.go
sturmf: Thanks for the comment, unfortunately I'm not sure what to do about that. I think with rsync it actually makes sense to pull. I would then do a rsync --delete src dst. Which would update the state of the source in sync. Unfortunately for that I would need the url in te Pull call, since it is not saved in the data directory as with git or other vcses.

How should I proceed in updating this branch?

vcs/rsync.go
sturmf: Tested it and all seems ok without the line. Therefore I updated the branch accordingly.

vcs/rsync.go
sturmf: Can you give me a few more hints how I could improve this PR to get it eventually merged?

PR #170, opened 2015-11-04
Delete and re-clone the repo if pull fails

cheister:
If a repo is forced pushed then a regular "git pull" will fail and hound will not index the latest repository contents.  This change should fix that issue by re-cloning if a pull operation fails.

### DIFF
--- a/searcher/searcher.go
+++ b/searcher/searcher.go
@@ -313,7 +313,7 @@ func updateAndReindex(
 	newRev, err := wd.PullOrClone(vcsDir, repo.Url)
 
 	if err != nil {
-		log.Printf("vcs pull error (%s - %s): %s", name, repo.Url, err)
+		log.Printf("vcs clone error (%s - %s): %s", name, repo.Url, err)
 		return rev, false
 	}
 
--- a/vcs/vcs.go
+++ b/vcs/vcs.go
@@ -71,7 +71,17 @@ func exists(path string) bool {
 // if the working directory is absent and pulling otherwise.
 func (w *WorkDir) PullOrClone(dir, url string) (string, error) {
 	if exists(dir) {
-		return w.Pull(dir)
+		newRev, err := w.Pull(dir)
+
+		if err == nil {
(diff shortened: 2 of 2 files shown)


kellegous:
I'm kind of worried about deleting on all pull failures. Etsy's hound instance, at times, has experienced a non-trivial number of transient connectivity issues. I'm trying to imagine what would have happened if we ended up deleting all the repos during these periods. I wonder if there is a more reliable way to separate a corrupt working copy from connectivity issues?

cheister:
We could return a network connectivity error message from Pull and not remove the directory in that case?  

Alternatively we could change the `git pull` operation to be a `git fetch --prune --no-tags --depth 1 origin +master:remotes/origin/master && git reset --hard origin/master` but that assumes the default branch for the repo is master.

mikepurvis:
> Alternatively we could change the git pull operation to be a git fetch --prune --no-tags --depth 1 origin +master:remotes/origin/master && git reset --hard origin/master but that assumes the default branch for the repo is master.

IMO this would be the best approach. #213 adds branch specifier— the logical next step there would be to use `git ls-remote` to figure out the branch in the case where it isn't explicitly given.

mikepurvis:
Even better might be to re-set the origin URL each time, in case it has changed:

```
git set-url origin %s
git fetch --prune --no-tags --depth 1 origin +master:remotes/origin/master
git reset --hard origin/master
```

mikepurvis:
Implemented in c21c82272279f72693588ee6dd70275b884f1ba0 if anyone's interested.

zstyblik:
@mikepurvis, has this issue been fixed then?

mikepurvis:
Not upstream— this repo is unmaintained AFAICT. You're welcome to use/examine the changes on my fork: https://github.com/mikepurvis/hound/commits/master

kellegous:
Closing since d99d1dbe667c24d615ab8ab13ccf1c093980a34a addresses this.

PR #173, opened 2015-11-20
Show matches count in results

francoisfreitag:
Displays matches count for each repo and each file in the repo.

### DIFF
--- a/ui/assets/css/hound.css
+++ b/ui/assets/css/hound.css
@@ -254,6 +254,10 @@ button:focus {
   background-color: #f5f5f5;
 }
 
+.file > .title > .matches-count {
+  color: #666;
+}
+
 .title a {
   color: #666;
 }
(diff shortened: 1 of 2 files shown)


jklein:
Seems reasonable, @kellegous take a look and merge when ready.

kellegous:
Sorry, Haven't been doing my job with these pull requests.

The repo level count is a little different from the file level. It's not sufficient to do `result.Matches.length` because the `Matches` array may be shortened due to a range request. You will notice that, by default, hound just returns the first 20 files matched in a repo. You will want to use `result.FilesWithMatches` instead, which will give you the true number of files that match (not just the ones that were returned).

egonSchiele:
FYI, this *almost* works correctly...there's a small mismatch: http://i.imgur.com/YEn7UfO.png
One place says "82 files", and the other one says "81 files".

Here's another example where it is "19 files" in one place but "27 files" in the other: http://i.imgur.com/6SN9aXn.png

PR #181, opened 2016-01-05
Added options to speficy defaults for poll frequence and vcs

pfrandsen:
Added options to speficy defaults for ms-between-polls and vcs in config.json - if any of the attributes are not defined the hardcoded values in the code (30000, "git") will be used as in the current version of the application.

Example config.json

{
    "dbpath" : "data",
    "repo-defaults" : {
        "ms-between-poll" : 30000,
        "vcs" : "git"
    },
    "repos" : {
        "Hound" : {
            "url" : "https://www.github.com/etsy/Hound.git"
        }
    }
}

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -33,6 +33,11 @@ type Repo struct {
 	EnablePushUpdates *bool          `json:"enable-push-updates"`
 }
 
+type RepoDefaults struct {
+	MsBetweenPolls int    `json:"ms-between-poll"`
+	Vcs            string `json:"vcs"`
+}
+
 // Used for interpreting the config value for fields that use *bool. If a value
 // is present, that value is returned. Otherwise, the default is returned.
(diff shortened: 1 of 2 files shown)


pfrandsen:
Added feature request https://github.com/etsy/hound/issues/185

bschlenk:
This should be extended to the base-url and anchor as well. It would be useful to define these globally if you work at a company with a custom git browsing website.

config/config.go
```diff
@@ -33,6 +33,11 @@ type Repo struct {
 	EnablePushUpdates *bool          `json:"enable-push-updates"`
 }
 
+type RepoDefaults struct {
```
sahildua2305: The name of the struct should be singular. So, we can have `RepoDefaultSetting` or `RepoDefault` here.

PR #189, opened 2016-02-02
fix isTextFile partial rune read

DmitryBochkarev:


DmitryBochkarev:
@kellegous @avleen can somebody review this one, please?

kellegous:
@DmitryBochkarev Can you give me some background on what you are fixing?

DmitryBochkarev:
We have valid utf-8 source code files, but they longer than 2048 bytes, so if on 2048 byte started new multibyte character, than current isTextFile function reading only first byte of it and of couse this is not valid utf-8 sequence.

kellegous:
Ah, I see. Yes, that makes sense.

index/index.go
```diff
@@ -241,7 +242,32 @@ func isTextFile(filename string) (bool, error) {
...
+	res, validSize := validUTF8(buf)
+	// if remaining non valid less than utf8.UTFMax probably we are read part of rune just assume it's valid
+	if n == filePeekSize && n-validSize < utf8.UTFMax {
+		res = true
+	}
+	return res, nil
+}
+
+// copy paste from go source https://golang.org/src/unicode/utf8/utf8.go?s=9677:9702#L387, with valid bytes size
+func validUTF8(p []byte) (bool, int) {
```
kellegous: Rather than rewriting `utf8.ValidUTF8`, is it possible to use `utf8.DecodeLastRune` to find the end of the last valid rune and then pass a slice of buf to `utf8.ValidUTF8`?

DmitryBochkarev: you are right, give me 5 minutes, to fix this

DmitryBochkarev: `utf8.DecodeLastRune` doesnt work this way :( it trying to parse from end of sequence but on end not valid character, so it's failed.

DmitryBochkarev: actually i think using custom validUTF8 function it is fastest way to check file is text or not

kellegous:
@DmitryBochkarev Can you verify that #190 works for you?

PR #190, opened 2016-02-04
Integrate pr/189

kellegous:


index/index.go
```diff
@@ -239,9 +240,33 @@ func isTextFile(filename string) (bool, error) {
...
+	for i < n {
+		if p[i] < utf8.RuneSelf {
+			i++
+		} else {
+			_, size := utf8.DecodeRune(p[i:])
+			if size == 1 {
+				// All valid runes of size 1 (those below RuneSelf) were handled above. This must be a RuneError.
+				// If we're encountering this error within UTFMax of the end and the current byte could be a
+				// valid start, we'll just ignore the assumed partial rune.
+				return n-i < utf8.UTFMax && utf8.RuneStart(p[i])
```
DmitryBochkarev: it's look nice but this check should occur only if `n == filePeekSize`
```go
func validUTF8IgnoringPartialTrailingRune(p []byte, ignoreTrailing bool) bool {
	i := 0
	n := len(p)

	for i < n {
		if p[i] < utf8.RuneSelf {
			i++
		} else {
			_, size := utf8.DecodeRune(p[i:])
			if size == 1 {
				// All valid runes of size 1 (those below RuneSelf) were handled above. This must be a RuneError.
				// If we're encountering this error within UTFMax of the end and the current byte could be a
				// valid start, we'll just ignore the assumed partial rune.
				return ignoreTrailing && n-i < utf8.UTFMax && utf8.RuneStart(p[i])
			}
			i += size
		}
	}
	return true
}
```

and in isTextFile:

```go
return validUTF8IgnoringPartialTrailingRune(buf[:n], n == filePeekSize), nil
```

DmitryBochkarev:
:+1:

index/index.go
kellegous: Good point.

PR #194, opened 2016-02-26
Dockerfile: Run houndd as unprivileged user

fbochu:
Being root in a container is like being root on the host machine[1]. Houndd
doesn't need root priviliges so it can be run as unprivileged user.

[1] https://stackoverflow.com/questions/19054029/security-of-docker-as-it-runs-as-root-user

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,10 @@ FROM golang
 
 COPY . /go/src/github.com/etsy/hound
 ONBUILD COPY config.json /hound/
+RUN adduser --uid 999 --gecos ,,, --disabled-password --home /hound hound
 RUN go-wrapper install github.com/etsy/hound/cmds/houndd
 
+USER hound
 EXPOSE 6080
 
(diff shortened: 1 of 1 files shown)


kellegous:
Thanks.

kellegous:
Oops. I spoke too soon. I'm getting permission denied errors in /hound. I probably need to revert this.

fbochu:
I suppose you have mounted one of your host directory in the Docker container. This will not work except if the user with uid = 999 of the host machine have read write and execute rights on the mounted folder.

kellegous:
This may be the right answer. I know a lot of people tend to use data-only containers for this stuff. The main thing I need is for the `Dockerfile` to work with the instructions given in the README and for it to be as simple as possible (the instructions can certainly be changed and are currently lacking). Sadly, it's currently not very simple at all and it's not clear to me if people are running hound in a local container with a mount to their host or if they are running it in a cloud container.

PR #195, opened 2016-02-26
Dockerfile: Use golang alpine as base image

fbochu:
Using `golang:alpine` allow to reduce build image size from 740.6 MB to
264 MB.


kellegous:
Looks like golang:alpine is missing `git`

fbochu:
I closing this PR, it was replaced by https://github.com/etsy/hound/pull/197

Issue #168, opened 2015-10-27
Add a way to request hound to do a graceful shutdown.


kellegous:
Killing a running hound instance is currently dangerous. If the process is terminated in the middle of an index, it can result in a corrupt index or a corrupt VCS directory. Since hound does not yet auto-recover from corruption, this can make automated restarts (like for upgrades) trickier.

Initial thought on this one is a signal handler that steals all the tokens from the index limiter before shutting down.

PR #198, opened 2016-02-27
Add graceful shutdown on SIGTERM

jnackman:
Addresses #168. This provides a way to explicitly shut down a searcher instead of stealing limiter tokens as suggested in the ticket. This may be useful for implementing config reloads or allowing searchers to perform cleanup work before shutting down.


kellegous:
Thank you. I have been wanting to add this and just haven't had the time. I'll try this out soon and merge it.

PR #210, opened 2016-05-23
Limit index with exclusion/inclusion patterns

imsut:
this change is to support "Add a per repo exclude_patterns directive to the config"
https://github.com/etsy/hound/issues/24

this also supports "include-patterns".

please let me know if you have any questions.


Wilfred:
I'd find this really useful too. @kellegous are you happy with this PR?

PR #213, opened 2016-06-06
Option to specify branch + bitbucket-specific URL/anchor defaults.

mikepurvis:
Spiritual successor to #35; allows specifying the tag or branch to index. Doesn't support hashes, as they can't be passed to `git clone --branch xxxx`.

The implementation here meets my needs, so I'm sharing it at this point, but potentially more work which would be required to merge this, roughly in order of importance:

  * branch-clone implementations for non-git VCSes.
  * enable using a default branch rather than `master`, when ref is unspecified.
  * for the case where the ref has changed from a previous run of Hound, the former clone should be blown away and recreated, as `git pull` is going to update the originally-cloned branch rather than switching (as discussed in #170).
  * a more elegant scheme for handling domain-specific defaults.
  * support for multiple branches (as was in #35).


ui/assets/js/common.js
```diff
@@ -5,7 +5,7 @@ var lib = {
 
     ExpandVars: function(template, values) {
       for (var name in values) {
-        template = template.replace('{' + name + '}', values[name]);
+        template = template.replace(new RegExp('{' + name + '}', 'g'), values[name]);
```
mikepurvis: Bitbucket's URL specifies the branch twice; this handles replacing multiple instances.

mikepurvis:
I'm making further changes to my master branch, so I'm going to withdraw this PR. Feel free to come and get the changes in the future if you want em.

dmsimard:

PR #219, opened 2016-08-23
fix #209: correctly decode the query parameters

### LINKED ISSUE #209: Hound doesn't consider "?q=foo+bar" equivalent to "?q=foo%20bar"
I'm fairly sure [this isn't standards compliant](http://stackoverflow.com/questions/22850844/html-form-is-converting-spaces-to-instead-of-20-when-building-url). In any case having an external web form that submits data to hound and searching for e.g. "foo bar" will result in "?q=foo+bar", which hound treats as a literal "+".

Instead it should treat "+" and "%20" as equivalent, and searching for a plus should be %2B.

JaderDias:
The '+' character is a space in application/x-www-form-urlencoded
content, such as the query part of a URL.

### DIFF
--- a/ui/assets/js/hound.js
+++ b/ui/assets/js/hound.js
@@ -60,7 +60,7 @@ var ParamsFromQueryString = function(qs, params) {
       return;
     }
 
-    params[decodeURIComponent(pair[0])] = decodeURIComponent(pair[1]);
+    params[decodeURIComponent(pair[0])] = decodeURIComponent(pair[1].replace(/\+/g, '%20'));
   });
 
 


sahildua2305:
Hi @JaderDias can you please submit a pull request with only the required changes excluding 85d1460f13ab87d7957a70539408f986b6e59838?

JaderDias:
@sahildua2305 
https://github.com/etsy/hound/commit/85d1460f13ab87d7957a70539408f986b6e59838 wasn't included by accident. It may not match the title of the pull request but it's a nice to have too. It was suggested by Ævar.

dgryski:
Avar's fix for the issues in https://github.com/etsy/hound/issues/249 should be a separate PR.

JaderDias:
Why is it so important for it to be in a separate PR? Both changes will improve the product.

dgryski:
To keep the PRs and patches handling a single trackable issue only.

JaderDias:
@sahildua2305  @dgryski  I finally removed Avar's commit from this pull request

dschott68:
Hello @JaderDias

Thanks so much for this submissions.  This should now be fixed by https://github.com/hound-search/hound/pull/321. Closing this out.

Cheers!

PR #221, opened 2016-09-15
Fixes #183.

### LINKED ISSUE #183: Don't stop indexing when users force-push to master
Currently, hound stops updating the checkout, just showing the error:

```
2016/01/12 11:51:09 vcs pull error (test-repo - http://git.internal/scm/data/test-repo.git): exit status 1
2016/01/12 11:51:39 Failed to git pull /hound/data/vcs-XXXXXXXXXXXXXXXX, see output below
error: Your local changes to the following files would be overwritten by merge:
        src/Test.java
Please, commit your changes or stash them before you can merge.
Aborting
```

Rather than doing a `git pull`, it would be better to do a `git fetch && git reset origin/master`.

kellegous:
Use more robust git combination that is tolerant to force pushed
branches.

### DIFF
--- a/vcs/git.go
+++ b/vcs/git.go
@@ -2,12 +2,15 @@ package vcs
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"log"
 	"os/exec"
 	"path/filepath"
 	"strings"
 )
+
+const defaultRef = "master"
 
 func init() {
 	Register(newGit, "git")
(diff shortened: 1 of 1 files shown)


vcs/git.go
```diff
@@ -44,12 +47,37 @@ func (g *GitDriver) HeadRev(dir string) (string, error) {
...
+		"origin",
+		fmt.Sprintf("+%s:remotes/origin/%s", defaultRef, defaultRef)); err != nil {
+		return "", err
+	}
+
+	if err := run("git reset", dir,
+		"git",
+		"reset",
+		"--hard",
+		fmt.Sprintf("origin/%s", defaultRef)); err != nil {
```
stanistan: Does this only every do `origin/master`? Do we support other branches in the repo config? If so, does this break that, because `git pull` takes care of that?

kellegous: There's only support for master right now. Though there have been feature requests to handle multiple branches, refs, etc. I just extracted the value to make it easier to add that branch support at a later date.

stanistan:
LGTM

PR #224, opened 2016-10-07
Add branching to git

mwitkow:
This allows users to customize the ref to use from the branch. For example:

```json
  "vcs": "git",
  "vcs-config": {
    "ref": "releases/9.0"
},
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .build
 .DS_Store
 *.exe
+.idea
--- a/vcs/git.go
+++ b/vcs/git.go
@@ -2,6 +2,7 @@ package vcs
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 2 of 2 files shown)


igorwwwwwwwwwwwwwwwwwwww:
Yes please!

kylej-ibm:
Please merge this feature

derimagia:
url-pattern config option should get {ref} option and the default option should use this:

https://github.com/improbable-io/hound/blob/c6c5459ffb60207e13c0d8cfaab2e7eed8658108/config/config.go#L16

sbergot:
@derimagia this would be a nice addition, but right now this you can still do this by hardcoding the reference in the url-pattern.

frankchdez:
@mwitkow consider adding `"--branch", g.Ref,` to the Clone method

dmsimard:

PR #230, opened 2016-12-14
OpenShift Template

jmferrer:
Added openshift template with persistent storage.

Remember to create config.json file in /data .

Example: oc new-app -f hound-openshift-template.json -p DATA_VOL_SIZE=10Gi,APPLICATION_HOSTNAME=hound.example.com

### DIFF
--- a/openshift/hound-openshift-template.json
+++ b/openshift/hound-openshift-template.json
@@ -1,0 +1,251 @@
+{
+  "kind": "Template",
+  "apiVersion": "v1",
+  "metadata": {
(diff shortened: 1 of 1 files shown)


mrtazz:
Hi! Thanks for the contribution! As we are not using Openshift I don't think it makes sense to maintain this config file in the repository. However if you want to host it in a repo yourself and link to it from the [Quickstart Guide](https://github.com/etsy/hound#quick-start-guide) in the README, I think that would be a great addition.

PR #236, opened 2017-02-02
ui: Add option to hide/show an entire repo

berdim99:
Add a simple UI to hide and re-show an entire repo.

After hacking this I found https://github.com/etsy/hound/pull/235 which solves the same issue, though a bit differently. Either one is good with me, I would just love to see this feature in hound.

screen shot 2017-02-02 at 9 17 21 am


jmphilli:
+1

gwindes:
+1 this looks great and much needed for us as we have hundreds of repos :)

russell-dot-js:
++1

PR #240, opened 2017-02-19
Remove lowercasing of repo names

bschlenk:
The lowercasing of repo names is entirely unnecessary, and only serves
to obfuscate the repo list in the ui and to complicate things for anyone 
trying to use the rest api. By keeping the repo names intact, there is no
longer a need to first call the repos endpoint in order to lookup an
actual name, if the name is the only piece of information that is needed.


kellegous:
Thanks. I'm pretty sure the lower-casing is a relic. I can't even remember what problem it originally addressed.

Issue #244, opened 2017-03-12
Please merge https://github.com/google/codesearch/pull/47


dgryski:
The bug in siftUp() is present in your vendored copy, which means that for large numbers of repositories the merge step can produce out-of-order indexes which will cause queries to fail.

PR #247, opened 2017-03-23
Fixes a codesearch bug in the index writer.

kellegous:
Addresses #244

### DIFF
--- a/codesearch/index/write.go
+++ b/codesearch/index/write.go
@@ -483,6 +483,7 @@ func (h *postHeap) siftUp(j int) {
 			break
 		}
 		ch[i], ch[j] = ch[j], ch[i]
+		j = i
 	}
 }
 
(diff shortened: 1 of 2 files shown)


codesearch/index/write.go
```diff
@@ -483,6 +483,7 @@ func (h *postHeap) siftUp(j int) {
 			break
 		}
 		ch[i], ch[j] = ch[j], ch[i]
+		j = i
```
stanistan: Can you explain what's happening here? I'm super unfamiliar with this :)

kellegous: When the indexer is at work, it keeps some of the data it's operating on in a heap (looks a lot like this https://en.wikipedia.org/wiki/Binary_heap#Heap_implementation). The heap is just keeping things in order. When you add something new to a heap, you put it in the next leaf node spot, then you keep looking at the parent nodes to see if they need to be swapped to maintain order. This implementation has a bug where it only checks the leaf node against the first parent. So there are some cases (when the heap is deeper than a single level) where the heap can have the wrong order. This should only happen in rare cases, on big indexes (I haven't look at the exact conditions that cause an error).

Basically my decision to try to merge this include:

 1. This indeed looks like a bug in a classic heap impl.
 2. Most other forks of the codesearch code also decided to merge this.
 3. The reporter was able to reproduce an error in a production environment.

PR #248, opened 2017-04-13
Branch support

havocbane:
I have added support to hound to be able to update cloned repositories to non-default branches using a new item in `config.json`. I have been using this locally for a little bit, but I am looking forward to any feedback you can provide. Thank you!

### DIFF
--- a/config-example.json
+++ b/config-example.json
@@ -14,9 +14,14 @@
             "url" : "https://www.example.com/foo/hg",
             "vcs" : "hg"
         },
+        "MercurialRepoWithBranch": {
+            "url" : "https://www.example.com/bar/hg",
+            "vcs" : "hg",
+            "branch" : "not-default"
+        },
         "Subversion" : {
             "url" : "http://my-svn.com/repo",
-            "url-pattern" : { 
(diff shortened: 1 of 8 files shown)


avar:
What VCS have you been running this with? Just looking at the code I don't see how it could work at all with git. The code currently does a reset --hard to origin/master: https://github.com/etsy/hound/blob/master/vcs/git.go#L76

So doesn't this just checkout some unrelated branch & then keep resetting it to whatever master is set to?

havocbane:
Thanks for the feedback! I have been using this strictly with mercurial repositories, so I can absolutely test this out with a git repository to make sure it's working as advertised. However, the way I have modified `PullOrClone` in `vcs.go` should cause the pull or clone to happen before a checkout/update. Granted, this isn't as efficient as it could be, so I can tweak that a little as well to avoid the hard reset.

havocbane:
I see one additional problem in a quick test against a git repo; the git clone command is passing a depth of 1 which causes the single-branch option to be implied. I will rework this a little and reopen. Thanks again.

dmsimard:

PR #256, opened 2017-07-24
Add test target to Makefile

jazzdan:
The README mentions that you can run `make test` to run the tests, but there is no test target in the Makefile. This commit adds that rule.

Is this the right way to run tests in Go?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,3 +11,6 @@ ui/bindata.go: .build/bin/go-bindata $(wildcard ui/assets/**/*)
 
 clean:
 	rm -rf .build
+
+test:
+	go test ./...


kellegous:
fwiw, this only works if the user has checked out hound into the proper spot in the GOPATH. I've struggled a little with knowing what to expect from user familiarity with go environments. I originally assumed that most contributors would be familiar with Go but that turned out to be not a great assumption. I would just look at the README and make sure that this works with the instructions for setup.

I have tried to continue to use these instructions to ensure they continued to work https://github.com/etsy/hound#requirements-1 and note that if you use that flow, the tests would be run from the Makefile in the tools directory.

So that's a lot of words to say, I think we should decide on one of two paths.
1 - Change the instructions to instruct developers that the need to put the project in their `GOPATH`.
2 - Move the test rule into the Makefile in tools.

I'm happy with either path, seems like more people understand 1 these days. But we should definitely be consistent with what's in the README.

sahildua2305:
> seems like more people understand 1 these days

I agree with this. I think almost all the golang projects these days use this kind of setup where the developers need to put the project in their `GOPATH` properly. So, I feel that it's a fair assumption to make that developers wanting to clone the repository and hack on it will know about `GOPATH` setup.

stanistan:
Yeah, the development instructions are correct in terms of linking to the `tools/Makefile` and then you can run `go test`. Also, running `make` from the root of the repo will generate different `bindata` than running from `tools` (via the README instructions). 

What is the expectation of having a root Makefile if we're telling people to use the one in `tools`?  *Is it better to delete the duplicate (root) Makefile*? Do we ever instruct people to use it?

sahildua2305:
Ok, so what's the purpose of `Makefile` in the root directory? Are we using it anywhere? @kellegous

kellegous:
@sahildua2305 That `Makefile` was there just to generate the `bindata.go` file. We kind of arrived at this state after a bit of a journey.

When hound was open sourced, it was built entirely with a Makefile in the root directory. You were not able to run `go get github.com/etsy/hound`. This had one big advantage, anyone wanting to work with the source just needed to run `make` regardless of whether they were working on the client or the server.

Folks familiar with Go were not happy with that solution as they expected to install hound with `go get`. I rearranged the repo to make that possible but that required that the `bindata.go` file be checked into the repo and so I created a `Makefile` in the root that just took care of that. There were a number of people at the time making changes to the client code and didn't touch (or understand) the Go code at all. When I made this change, I switched over to a "standard" Go development environment where you had to put the code into your `GOPATH`.

We continued to have confusion about how to clone hound and make changes after that round of changes. Folks who wanted to change the client code often had issues getting the server built. That's where the `tools` scripts came from. They were an attempt to have a short README that would get people up and running.

In my opinion, the `Makefile` we started with was the superior option in that it gave you a very clear path to get the source built into a runnable binary. But I understand that folks in the Go community are keen on `go get`.

What I'm thinking we should do now, is
 1. Update the `README.md` with steps to clone hound into a `GOPATH`.
 2. Merge the two `Makefile`s and just have the single Makefile assume that the source is on the `GOPATH`. (Maybe there is a way to double check that the code really is in the right directory).

sahildua2305:
Nice! Thanks for giving a complete overview of how things changed with time. I agree with the changes you are suggesting.

@jazzdan would you be interested in making the changes @kellegous suggested above. Anyway, I'll open a new issue for this.

jazzdan:
@sahildua2305 I'll try to take a pass at that this weekend. Thanks for all the info everyone!

PR #259, opened 2017-09-19
improve parsing of fully-protocoled URLs

### LINKED ISSUE #193: url-pattern: {url} - url:port is converted to url/port
When git url contains a port 
eg: ssh://git@stash.example.com:7999/foo/bar.git 

the url is converted to http://stash.example.com/7999

Roguelazer:
fixes #193, I think


Crocmagnon:
Any update on this?

salemhilal:
@Crocmagnon @Roguelazer yes, actually! We've been triaging issues as of late, but this would be an appreciated contribution.

We've recently fleshed out how JS test framework. Tests would be appreciated before we can merge this, particularly since regular expressions can be tricky to reason about.

PR #267, opened 2017-12-17
Fix xml parsing of open_search.xml

### LINKED ISSUE #239: open_search.xml malformed
The templating seems to substitute the `<` in the xml version declaration with `&lt;`. As a result, the search provider cannot be added in firefox.

Example: https://search.nix.gsc.io/open_search.xml

Adding the search provider fails in firefox with the cryptic error message
>Firefox could not download the search plugin from:
>https://search.nix.gsc.io/open_search.xml

sahildua2305:
This turned out to be more cumbersome than I had expected it to be. Below
is the description of what I did and the rationale behind it.

open_search.xml file was being served using html/template package and
since it doesn't know about xml contexts, it changes the first character
of xml template "<" into "\&lt;".

Now, we use two different packages:
- html/template - for parsing html template files
- text/template - for parsing text or xml template files

Here it's arguable that we could use encoding/xml package to parse the
xml template files in a better way. However, to keep the parsing logic
consistent, we are fine with using text/template only. If, in future, we
come across a use case where we want to parse and serve xml file where
we need to know the xml context while parsing the template, we may think
of switching to encoding/xml at the cost of inconsistency. There's no
need as of now to use it, because it's a very simple xml file.

Fixes #239 
/cc @stanistan

### DIFF
--- a/ui/content.go
+++ b/ui/content.go
@@ -1,9 +1,6 @@
 package ui
 
-import (
-	htemplate "html/template"
-	ttemplate "text/template"
-)
+import "io"
 
 // Current versions of some dependencies.
 const (
@@ -12,6 +9,15 @@ const (
 )
 
 var contents map[string]*content
+
+// This interface abstracts the Execute method on template which is
+// structurally similar in both html/template and text/template.
+// We need to use an interface instead of a direct template field
+// because then we will need two different fields for html template
+// and text template.
+type renderer interface {
+	Execute(w io.Writer, data interface{}) error
+}
 
 type content struct {
 
@@ -24,13 +30,8 @@ type content struct {
 	// The JavaScript sources used in this HTML page
 	sources []string
 
-	// This is only created in prd-mode, the pre-parsed template
-	// HTML template - used for serving HTML content
-	htpl *htemplate.Template
-
-	// This is only created in prd-mode, the pre-parsed template
-	// Text template - currently used for /open_search.xml only
-	ttpl *ttemplate.Template
+	// The parsed template - can be of html/template or text/template type
+	tpl renderer
 
 	// This is used to determine if a template is to be parsed as text or html
 	tplType string
--- a/ui/content.go
+++ b/ui/content.go
@@ -1,6 +1,9 @@
 package ui
 
-import "html/template"
+import (
+	htemplate "html/template"
+	ttemplate "text/template"
+)
 
 // Current versions of some dependencies.
 const (
@@ -22,7 +25,15 @@ type content struct {
 	sources []string
 
 	// This is only created in prd-mode, the pre-parsed template
-	tpl *template.Template
+	// HTML template - used for serving HTML content
+	htpl *htemplate.Template
+
+	// This is only created in prd-mode, the pre-parsed template
+	// Text template - currently used for /open_search.xml only
+	ttpl *ttemplate.Template
+
+	// This is used to determine if a template is to be parsed as text or html
+	tplType string
 }
 
 func init() {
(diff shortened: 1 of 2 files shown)


ui/content.go
```diff
@@ -22,7 +25,15 @@ type content struct {
 	sources []string
 
 	// This is only created in prd-mode, the pre-parsed template
-	tpl *template.Template
+	// HTML template - used for serving HTML content
+	htpl *htemplate.Template
+
+	// This is only created in prd-mode, the pre-parsed template
+	// Text template - currently used for /open_search.xml only
+	ttpl *ttemplate.Template
```
kellegous: Seems like you should be able to better take advantage of the fact that `html/template.Template.Execute` and `text/template.Template.Execute` are structurally equivalent. You should be able to store both types into an interface field.
```
type renderer interface {
    Execute(w io.Writer, data interface{}) error
}
```
That would at least clean up one set of switch statements for dispatching.

ui/ui.go
```diff
@@ -133,6 +149,14 @@ func (h *prdHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 // Renders a templated asset in prd-mode. This strategy will embed
 // the sources directly in a script tag on the templated page.
 func renderForPrd(w io.Writer, c *content, cfgJson string, r *http.Request) error {
+	// If the requested asset is of type xml or text, short cut the execution
+	// using text/template package.
+	// See - https://github.com/etsy/hound/issues/239
+	if c.tplType == "xml" || c.tplType == "text" {
+		return c.ttpl.Execute(w, map[string]interface{}{
+			"Host": r.Host,
```
kellegous: Is there any reason not to pass the same data map to all templates? If not, seems like consistency would help us avoid surprises in future additions.

sahildua2305: No reason as such. However, you're right. Inconsistency may hurt in future additions. I'll fix that.

ui/content.go
sahildua2305: Thanks @kellegous, I was thinking of using an interface in this case, but wasn't able to figure out how to use one. I'll try again and push the fixes.

sahildua2305:
@kellegous I just pushed the changes you suggested and the code is much simpler now. Thanks for suggestions! 🙂

ui/ui.go
```diff
@@ -2,8 +2,10 @@ package ui
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
-	"html/template"
+	htemplate "html/template"
```
stanistan: Rename these to `html_template` and `text_template`

ui/ui.go
```diff
@@ -185,9 +201,24 @@ func newPrdHandler(cfg *config.Config) (http.Handler, error) {
 			return nil, err
 		}
 
-		cnt.tpl, err = template.New(cnt.template).Parse(string(a))
-		if err != nil {
-			return nil, err
+		// For more context, see: https://github.com/etsy/hound/issues/239
```
stanistan: Dry this bit up into a function :)

sahildua2305: It's actually different for prod and dev.

ui/ui.go
sahildua2305: Good one! Will do.

ui/ui.go
sahildua2305: In prod, we load the contents of template from bindata, however, in dev mode, we load the contents of template directly from the file.

stanistan:
Thanks @sahildua2305!

sahildua2305:
Thanks @stanistan and @kellegous :)

PR #273, opened 2017-12-30
golang version: remove 1.3 support

sahildua2305:
Remove golang 1.3 from .travis.yml and also mention the support of 1.4+
golang in requirements.

cc @kellegous


sahildua2305:
There we go, @kellegous! 😃

PR #283, opened 2018-05-04
Fix inconsistent sorting of search results

Nikerabbit:
Fixes https://github.com/etsy/hound/issues/281


kellegous:
The commit was merged with #290

PR #291, opened 2018-07-20
Adds webpack-dev-server to houndd --dev

danielmendel:
The current `houndd --dev` setup for assets is broken because it loads sourcecode directly and chokes on the recently added ES6 import statement.

This change spawns [webpack-dev-server](https://github.com/webpack/webpack-dev-server) when running `houndd --dev` and points the `<script>` tags at that host for the built assets. 

It appears that the webpack server is fully cleaned up when the process exits, I wasn't sure if I was supposed to do something in the shutdown function ʅʕ•ᴥ•ʔʃ

### DIFF
--- a/cmds/houndd/main.go
+++ b/cmds/houndd/main.go
@@ -6,7 +6,9 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"os/exec"
 	"os/signal"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"syscall"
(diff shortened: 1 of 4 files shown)


kellegous:
Sorry for the slow review. I set this up locally today and realized that without some additional tweaking, I get:

```
2018/08/19 18:42:10 [DEV] starting webpack-dev-server at localhost:8080...
2018/08/19 18:42:10 fork/exec ./node_modules/.bin/webpack-dev-server: no such file or directory
2018/08/19 18:42:10 running server at http://localhost:6080...
```

Thinking about the best way to ensure we have a populated `node_modules`.

kellegous:
Ah ha, my plan is to just update the README.md for dev mode to instruct folks to run `npm install`

PR #312, opened 2019-06-16
Updated build instructions post-Etsy

jacobrose:
The git repo path has changed now that Hound is its own org.

I also added an example build session for folks like me who
don't have other Go projects, and included the "go get" step
in these new build instructions, without which I wasn't able
to build.


jacobrose:
Whoops, I tried something to send y'all two PRs at the same time, and screwed up this PR. I'll re-do it.

PR #315, opened 2019-06-16
Updated build instructions post-Etsy

jacobrose:
The git repo path has changed now that Hound is its own org.

I also added an example build session for folks like me who
don't have other Go projects, and included the "go get" step
in these new build instructions, without which I wasn't able
to build.


kellegous:
I rebased this, updated the instructions a little bit and merged it as a part of #320.

PR #327, opened 2019-11-17
Remove jQuery

elchininet:
In the project, jQuery is used only for Ajax requests and for merging objects. There is no need to load this whole library for these operations.

There are popular libraries that can be used specifically for those purposes ([reqwest](https://www.npmjs.com/package/reqwest) and [merge-anything](https://www.npmjs.com/package/merge-anything)).

This makes the process of updating a library easier because no Go code is needed to be changed, only the version of the packages in the package.json should be updated.

This pull request assumes that [this commit](https://github.com/elchininet/hound/commit/89266b7a65416997ecdf669df585878076ba3391) is already merged.


yelizariev:
Do you have some statistics about performance improvement after this update?

elchininet:
> Do you have some statistics about performance improvement after this update?

If there is an improvement in performance, it should not be huge (I have not measure it). The improvement here is more related to maintainability and upgrading, if one wants to update the library, instead of updating the `jQuery` file, the `ui/bindata.go` file, and the `ui/content.go` file, one just needs to deal with the `npm` packages.

twizzyyanki:
hi @elchininet thanks for making this change. Since it's been a while I couldn't merge your request so I recreated it in [this pull request](https://github.com/hound-search/hound/pull/465) and merged it. I'll continue from where you stopped on removing `merge-anything` and `create-react-class`.

elchininet:
Thanks for the update @twizzyyanki,
As it has been a long time, I don‘t remember the order of the branches but I did a lot of improvements in my fork. If you dig a bit you can find some of them interesting.
Regards

PR #344, opened 2020-07-09
Add fake VCS driver

tgulacsi:
Add a fake VCS driver, to allow indexing a local, not versioned directory.

### DIFF
--- a/vcs/fake.go
+++ b/vcs/fake.go
@@ -3,11 +3,8 @@ package vcs
 import (
 	"crypto/sha1"
 	"fmt"
-	"io/ioutil"
-	"log"
 	"os"
 	"os/exec"
-	"path/filepath"
 	"strings"
 )
 
@@ -26,7 +23,7 @@ func (g *FakeDriver) HeadRev(dir string) (string, error) {
 	cmd := exec.Command(
 		"tar",
 		"cf", "-",
-		".",
+		"./",
 	)
 	cmd.Dir = dir
 	cmd.Stdout = hsh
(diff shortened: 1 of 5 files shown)


vcs/fake.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (g *FakeDriver) HeadRev(dir string) (string, error) {
+	hsh := sha1.New()
+	cmd := exec.Command(
+		"tar",
+		"cf", "-",
+		"./",
+	)
+	cmd.Dir = dir
+	cmd.Stdout = hsh
```
salemhilal, line 29: Clever! I really like that this doesn't leave tar files in the filesystem and doesn't (or shouldn't) be a problem with large inputs.

vcs/fake.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+const fakeSpec = ".source"
+
+func (g *FakeDriver) Pull(dir string) (string, error) {
+	return g.HeadRev(dir)
+}
+
+func (g *FakeDriver) Clone(dir, url string) (string, error) {
+	src := strings.TrimPrefix(url, "file://")
```
salemhilal, line 44: Is it worth adding a check at initialization time to ensure that the `url` config starts with `file://`?

vcs/fake.go
```diff
@@ -0,0 +1,53 @@
...
+	cmd.Dir = dir
+	cmd.Stdout = hsh
+	if err := cmd.Run(); err != nil {
+		return "", fmt.Errorf("%v: %+v", cmd.Args, err)
+	}
+
+	return fmt.Sprintf("%x", hsh.Sum(nil)), nil
+}
+
+const fakeSpec = ".source"
```
salemhilal, line 37: Is this used anywhere?

salemhilal (changes requested):
I'm really into this approach! To clarify my understanding:

1. The new driver imports a local directory as a repository by creating a symlink to it
2. The index crawler is updated to follow symlinks
3. A reference hash is approximated by hashing the tar'd output of the directory

If that all sounds good, I think these are the few small things to do before I can merge it:

- [ ] Potentially consider "none" as the config name and the name of the driver.
- [ ] Validate that the driver's initialized with a url that begins with `file://`, and return an error from `newFake()` otherwise. 
- [ ] Add a test file. We don't have any strong integration test patterns yet, but at least validating ^ logic would be a start. It would be extra nice to test the hashing logic as well. I'm also very open to integration test thoughts, if you have any.
- [ ] Add doc blocks and delete that `const fakeSpec` line (I _think_ it isn't used anywhere).

salemhilal:
Out of curiosity, are you running into a situation that's preventing a `git init` in a local directory?

mrexodia:
It would be really great if this could be merged. Is it okay if I take this over? I think `"vcs": "local"` or `"vcs": "file"` would be more descriptive than `none` though.

salemhilal:
@mrexodia feel free to take this over! I'd also love to know more about your use case for having code search without some sort of version control.

mrexodia:
Hopefully will be able to get to it this month... My use case is indexing a few GB of leaked source code and adding things to git would take a lot of disk space/time.

salemhilal:
Hey @mrexodia, I was under the impression that you'd be taking this PR over. Are you able to resolve the outstanding items and comments in this PR?

mrexodia:
Hopefully eventually, don’t have the most time in the world but I’ll get to it…

PR #345, opened 2020-07-11
Git VCS driver: Support non-default target refs

rwliang:
Currently, only the default ref (`master`) is supported.

- Allow target non-default target refs to be specified via configs.
- Add config-driven target ref auto-detection using the `HEAD` branch.
- Update default repo base URL pattern to link to the content revision,
  accounting for potential non-default target ref usage.
- Add global-level VCS configs which are merged into the repo-level configs
  during startup (repo-level vals take precedence).

### DIFF
--- a/config-example.json
+++ b/config-example.json
@@ -3,6 +3,11 @@
     "dbpath" : "data",
     "title" : "Hound",
     "health-check-uri" : "/healthz",
+    "vcs-config" : {
+        "git": {
+            "detect-ref" : true
+        }
+    },
     "repos" : {
         "SomeGitRepo" : {
             "url" : "https://www.github.com/YourOrganization/RepoOne.git"
@@ -12,13 +17,19 @@
             "ms-between-poll": 10000,
             "exclude-dot-files": true
         },
+        "GitRepoWithDetectRefDisabled" : {
+            "url" : "https://www.github.com/YourOrganization/RepoOne.git",
+            "vcs-config" : {
+                "detect-ref" : false
+            }
+        },
         "SomeMercurialRepo" : {
             "url" : "https://www.example.com/foo/hg",
             "vcs" : "hg"
         },
         "Subversion" : {
             "url" : "http://my-svn.com/repo",
-            "url-pattern" : { 
+            "url-pattern" : {
                 "base-url" : "{url}/{path}{anchor}"
             },
             "vcs" : "svn"
@@ -49,6 +60,9 @@
             "url-pattern" : {
                 "base-url" : "https://{url}/src/master/{path}{anchor}",
                 "anchor" : "#{filename}-{line}"
+            },
+            "vcs-config" : {
+                "ref" : "main"
             }
         },
         "RepoWithPollingDisabled" : {
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"encoding/json"
 	"path/filepath"
 	"runtime"
 	"testing"
@@ -34,4 +35,21 @@ func TestExampleConfigsAreValid(t *testing.T) {
 			t.Fatal(err)
 		}
 	}
+
+	// Ensure that global VCS config vals are merged
+	repo := cfg.Repos["SomeGitRepo"]
+	vcsConfigBytes := repo.VcsConfig()
+	var vcsConfigVals map[string]interface{}
+	json.Unmarshal(vcsConfigBytes, &vcsConfigVals)
+	if detectRef, ok := vcsConfigVals["detect-ref"]; !ok || !detectRef.(bool) {
+		t.Error("global detectRef vcs config setting not set for repo")
+	}
+
+	repo = cfg.Repos["GitRepoWithDetectRefDisabled"]
+	vcsConfigBytes = repo.VcsConfig()
+	json.Unmarshal(vcsConfigBytes, &vcsConfigVals)
+	if detectRef, ok := vcsConfigVals["detect-ref"]; !ok || detectRef.(bool) {
+		t.Error("global detectRef vcs config setting not overriden by repo-level setting")
+	}
+
 }
--- a/vcs/git.go
+++ b/vcs/git.go
@@ -2,24 +2,49 @@ package vcs
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
 	"log"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 )
 
 const defaultRef = "master"
 
+var headBranchRegexp = regexp.MustCompile(`HEAD branch: (?P<branch>.+)`)
+
 func init() {
 	Register(newGit, "git")
 }
 
-type GitDriver struct{}
+type GitDriver struct {
+	DetectRef     bool   `json:"detect-ref"`
+	Ref           string `json:"ref"`
+	refDetetector refDetetector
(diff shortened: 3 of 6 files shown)


salemhilal:
Hey, just so that you're not left hanging, we're putting eyes on this PR next.

salemhilal:
I'm still looking through this PR (nice work!) and one thing I've noticed is that this doesn't seem to work for local repositories. 
I used the config below, which ended up failing to index my local folder. 

```json
{
    "max-concurrent-indexers" : 2,
    "dbpath" : "data",
    "title" : "Hound",
    "health-check-uri" : "/healthz",
    "repos" : {
        "Hound" : {
            "url" : "https://www.github.com/hound-search/hound.git"
        },
        "LocalFolder" : {
            "url" : "file:///Users/salemhilal/Code/some-local-repo",
            "vcs-config": {
                "detect-ref": false,
                "ref": "main"
            }
        }
    }
}
```

Specifically, it looks like it's trying to still use `master` as the ref, and it assumes it has a remote.
```
2020/07/24 15:05:35 Searcher started for Hound
2020/07/24 15:05:35 Searcher started for LocalFolder
2020/07/24 15:05:35 Failed to git fetch /Users/salemhilal/go/src/github.com/hound-search/hound/data/vcs-553b942fb36f8d5521e6625aa4157f26bd2e954c, see output below
fatal: Couldn't find remote ref master
fatal: the remote end hung up unexpectedly
Continuing...
```

This seems to be a bug in hound currently, but it'd be nice to support non-default target refs everywhere. I would dig into it a bit more but I'm about to be out for the weekend. I'm happy to look into this further when I'm back (mid next week). 

Also, apologies in advance if I'm missing something obvious here. I'm still ramping up on the details of this codebase.

rwliang:
Hey @salemhilal - thanks for taking a look! 🙏. 

Hmm local repo indexing seems to be working for me with this config:
```json
{
  "max-concurrent-indexers" : 2,
  "dbpath" : "data",
  "title" : "Hound",
  "health-check-uri" : "/healthz",
  "repos" : {
    "Hound" : {
      "url" : "https://github.com/etsy/hound.git"
    },
    "LocalFolder": {
      "url" : "file:///Users/rliang/local-rep",
      "vcs-config" : {
        "detect-ref" : false,
        "ref" : "main"
      }
    }
  }
}
```
`master` shouldn't be used in the fetch due to the explicit target ref config entry. Is it possible that the app was compiled without these changes applied?

salemhilal:
Hey, that’s about where I ended up before I had to leave. I thought I had the right branch of your fork checked out, but if you can’t reproduce it that sounds like something wrong on my end. I’m unfortunately out the door already, but I’ll look at it again more closely on Monday.

salemhilal:
Ok! I'm home again and can confirm that this is working as documented (I had been using `go install github.com/rwliang/hound/cmds/houndd` rather than setting your repo as another origin and using just `go install github.com/hound-search/hound/cmds/houndd`) . I'm going to put some closer eyes on the PR today and get one of the other maintainers to look, but this looks good to me.

vcs/git.go
```diff
@@ -47,37 +62,79 @@ func (g *GitDriver) HeadRev(dir string) (string, error) {
...
-	if err := run("git fetch", dir,
+	var targetRef string
+	if len(g.Ref) > 0 {
+		targetRef = g.Ref
+	} else if g.DetectRef {
+		targetRef = detectRef(dir)
+	}
+	if len(targetRef) == 0 {
+		targetRef = defaultRef
+	}
```
salemhilal, line 120: Would you be willing to add some tests for this logic? Having something that enforces the precedence that these rules are evaluated could be useful.

vcs/git.go
```diff
@@ -47,37 +62,79 @@ func (g *GitDriver) HeadRev(dir string) (string, error) {
...
 	}
-	return nil
+
+	return string(out), nil
 }
 
 func (g *GitDriver) Pull(dir string) (string, error) {
-	if err := run("git fetch", dir,
+	var targetRef string
+	if len(g.Ref) > 0 {
```
kellegous, line 113: For consistency, empty string comparisons should be `g.Ref == ""`

vcs/git.go
```diff
@@ -99,7 +156,7 @@ func (g *GitDriver) Clone(dir, url string) (string, error) {
 		return "", err
 	}
 
-	return g.HeadRev(dir)
+	return g.Pull(dir)
```
salemhilal, line 159: Can you give me more details on what this additional `Pull` does here?

salemhilal:
As a note, all tests are currently passing.

salemhilal:
Hey @rwliang, I just wanted to bump this PR because it sounds like it'll unblock some additional work. If you're able to take care of those last handful of comments, that'd be incredibly appreciated. If not, one of us will try and address them. Thanks again either way!

rwliang:
Hey @salemhilal - really sorry for the delay. I should be able to address the feedback sometime in the next two days 🙏.

vcs/git.go
rwliang: Will do - thanks!

config-example.json
```diff
@@ -12,13 +17,19 @@
             "ms-between-poll": 10000,
             "exclude-dot-files": true
         },
+        "GitRepoWithDetectRefDisabled" : {
+            "url" : "https://www.github.com/YourOrganization/RepoOne.git",
+            "vcs-config" : {
+                "detect-ref" : false
+            }
+        },
```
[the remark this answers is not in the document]
rwliang: Will do - thanks!

config/config_test.go
```diff
@@ -34,4 +35,21 @@ func TestExampleConfigsAreValid(t *testing.T) {
...
 		}
 	}
+
+	// Ensure that global VCS config vals are merged
+	repo := cfg.Repos["SomeGitRepo"]
+	vcsConfigBytes := repo.VcsConfig()
+	var vcsConfigVals map[string]interface{}
+	json.Unmarshal(vcsConfigBytes, &vcsConfigVals)
+	if detectRef, ok := vcsConfigVals["detect-ref"]; !ok || !detectRef.(bool) {
+		t.Error("global detectRef vcs config setting not set for repo")
```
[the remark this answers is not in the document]
rwliang: Hmm the `vcsConfigVals` in the next block are pulled from another repo config and are used in a separate assertion. I think we'd want to proceed to capture all the errors in one go.

vcs/git.go
rwliang: Will do - thanks!

vcs/git.go
rwliang: `Clone` initially checks out the default HEAD branch. The `Pull` ensures that we check out out the target ref.

rwliang:
Hey folks - attempted to address the feedback. Apologies for the delays and thanks again for the reviews!

vcs/git.go
salemhilal: got it, much appreciated.

salemhilal:
We have a meeting tomorrow, so with any luck we'll be able to merge this PR this week. Thanks again @rwliang!

salemhilal:
Ok, we've tested this internally with a repo that's been updated to use `main`. Everything works as expected.

salemhilal:
@rwliang Thank you so, so much for this PR. It is hugely appreciated!

rwliang:
Thanks, folks! 🙏

PR #223, opened 2016-10-02
Add a search option to exclude file paths with a regexp

vrischmann:
This PR adds a search field to exclude file paths with a regexp. It works exactly like the current file path search, except if it matches it rejects the file.

I made this because I have a lot of Go repositories with vendored dependencies and I'd like to be able to ignore all those files when searching.

PR #347, opened 2020-08-11
Exclude file path

vrischmann:
This PR adds a search field to exclude file paths with a regexp. It works exactly like the current file path search, except if it matches it rejects the file.

This is the same as #223 but because I had removed my hound fork a long time ago I'm unable to update that PR.

I rebased on the latest master and did a quick test that it still works.

### DIFF
--- a/index/index.go
+++ b/index/index.go
@@ -41,11 +41,12 @@ type IndexOptions struct {
 }
 
 type SearchOptions struct {
-	IgnoreCase     bool
-	LinesOfContext uint
-	FileRegexp     string
-	Offset         int
-	Limit          int
+	IgnoreCase        bool
+	LinesOfContext    uint
+	FileRegexp        string
+	ExcludeFileRegexp string
+	Offset            int
+	Limit             int
 }
 
 type Match struct {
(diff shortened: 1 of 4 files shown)


index/index.go
```diff
@@ -167,6 +168,14 @@ func (n *Index) Search(pat string, opt *SearchOptions) (*SearchResponse, error)
 		}
 	}
 
+	var efre *regexp.Regexp
```
jacobrose, line 171: I think our variable names are getting a tad too cryptic -- what do you think about `excludeFre`?

jacobrose (commented):
Hi, @vrischmann! Great contribution, one which we've had requests for. I left a comment for you to consider including in your PR, but we're ready to approve either way. Please let me know if you'd like to make that change before merging. Thanks again!

index/index.go
vrischmann: Sure, I'll rename it.

jacobrose (approved):
Thank you so much, @vrischmann!

ui/assets/js/hound.js
```diff
@@ -373,6 +374,23 @@ var SearchBar = React.createClass({
   filesGotFocus: function(event) {
     this.showAdvanced();
   },
+  excludeFilesGotKeydown: function(event) {
+    switch (event.keyCode) {
+    case 38:
+      // if advanced is empty, close it up.
+      if (this.refs.excludeFiles.getDOMNode().value.trim() === '') {
```
salemhilal, line 381: I think this check made sense when there was only one text field, but now that there's two, you'll likely have to check that both this text field and the `this.refs.files` text field are empty or you'll end up closing the advanced view when it's still in use.

salemhilal (commented):
Sorry for the slow back-and-forth. I have one comment, but this looks great (and super straightforward).

ui/assets/js/hound.js
vrischmann: Okay, I addressed that in the latest commit.

salemhilal (approved):
Wicked fast, thank you!

salemhilal:
Gave this branch one last local test, and everything works as expected.
Thanks again for reviving such an old branch @vrischmann. Your work is super appreciated.

PR #341, opened 2020-02-27
Maintenance

klingtnet:
This pull request does a bit of project maintenance:

# CI Improvements

- Travis CI is replaced by Github Actions
- Go runtimes are updated, tests run against Go 1.12-Go 1.14
- Tests are now run against Windows and Linux
- The code is now linted

# Project improvements

- hound is now a Go modules project
- `go-bindata` dependency is updated and replaced by a maintained fork
- all linter errors are now fixed (removal of unused code, unchecked errors¹)

¹ `unmmap` fails in the `write` unit tests if the error is propagated. Not sure if this is a problem outside the test environment. Fixing this is outside the scope of this PR.

PR #353, opened 2020-10-15
Initialise go modules

blobbered:

This PR is responsible for introducing [Go modules](https://blog.golang.org/using-go-modules) as hound's dependency management.


salemhilal (changes requested):
This seems to work as expected locally, and the installation instructions didn't even have to change! 

One thing to update: this change looks like it bumps the minimum version of Go that Hound relies on to 1.13, so that will need to be updated in the `README.md`. Aside from that, this PR looks good to me.

blobbered:
will do!

salemhilal:
Nice, thank you!

PR #354, opened 2020-10-15
Setup Github Actions CI

blobbered:

Original description:
```
Replaces Travis CI with Github Actions for better integration.
Jobs are run against more recent Go versions, namely 1.12 to 1.14 .
 .
For further documentation see

- https://github.com/actions/setup-go
- and https://help.github.com/en/actions

This adds `-race` flag to the Go tests and will run a meta linter
(golangci-lint).
```


README.md
```diff
@@ -1,6 +1,6 @@
 # Hound
 
-[![Build Status](https://travis-ci.org/hound-search/hound.svg?branch=master)](https://travis-ci.org/hound-search/hound) 
+![.github/workflows/go.yaml](https://github.com/spreadshirt/hound/workflows/.github/workflows/go.yaml/badge.svg)
```
salemhilal, line 3: I think this isn't quite the right URL for our badge. This points to the badge that lives in a forked copy of Hound.

salemhilal (commented):
For now, can you remove the deletion of the `travis.yml`, and the updated badge? I want to make sure tests work as expected with github workflows before tearing travis out (for example, I don't totally know how this would work with go v1.12, since we use go modules).

salemhilal:
Sorry for the long back-and-forth on this one. It looks like most of the tests are failing as-written. You can see them in your fork under the "Actions" tab. Here's the most recent job:
https://github.com/blobbered/hound/actions/runs/329382320/workflow

Based on this issue, I think these problems are because the repo isn't a go module in your branch:
https://stackoverflow.com/questions/62945517/github-action-golang-cannot-find-package

The commit to add go module support to Hound has been merged (thank you for that!), so I think updating this branch should fix the problem. If the branch is up to date and the build is passing, I'll be down to merge this in.

blobbered:
You're right! A merge helped and I've tidied the mods file.

As a side note- commit `e910b07` removes the lint command, since the code is not yet linted and the command was failing. I intend to revert it in the [vacuum lint PR](https://github.com/hound-search/hound/pull/355) so that we can keep these PRs relatively separate, assuming that this PR comes first and then the linter PR comes next.

salemhilal:
I see a passing test in the actions tab in your fork! I noticed you removed `go-bindata` from `go.mod`, but we do in fact use that library. If you put it back in, I think I can merge this PR.

blobbered:
ah, okay! will do

go.sum
```diff
@@ -1,6 +1,2 @@
-github.com/blang/semver v1.1.0 h1:ol1rO7QQB5uy7umSNV7VAmLugfLRD+17sYJujRNYPhg=
 github.com/blang/semver v3.5.1+incompatible h1:cQNTCjp13qL8KC3Nbxr/y2Bqb63oX6wdnnjpJbkM4JQ=
 github.com/blang/semver v3.5.1+incompatible/go.mod h1:kRBLl5iJ+tD4TcOOxsy/0fnwebNt5EWlYSAyrTnjyyk=
-github.com/go-bindata/go-bindata v1.0.0 h1:DZ34txDXWn1DyWa+vQf7V9ANc2ILTtrEjtlsdJRF26M=
-github.com/go-bindata/go-bindata v3.1.2+incompatible h1:5vjJMVhowQdPzjE1LdxyFF7YFTXg5IgGVW4gBr5IbvE=
-github.com/go-bindata/go-bindata v3.1.2+incompatible/go.mod h1:xK8Dsgwmeed+BBsSy2XTopBn/8uK2HWuGSnA11C3Joo=
```
salemhilal, line 6: Sorry for all the back-and-forth: can you revert the changes to this file as well?

blobbered: yeah! out of curiousity, how come go-bindata is an implicit dependency?

salemhilal: I don't have a great answer because I'm still relatively new to go modules, but I believe it's because `go-bindata` is only used in the Makefile and not directly in any go source files.

salemhilal:
Nice! [It looks like this config is working](https://github.com/blobbered/hound/actions/runs/356254325) so I'm going to merge this in. Thanks for all your patience with me and all these back-and-forths.

PR #356, opened 2020-10-15
Test windows

blobbered:

This PR is responsible for cleaning up the go build and test so that they run on windows.


dschott68:
Hi @blobbered 
It looks like [https://github.com/hound-search/hound/pull/356/commits/9617631d2b7c2c5751050066acde8e1a8eb83fc4](https://github.com/hound-search/hound/pull/356/commits/9617631d2b7c2c5751050066acde8e1a8eb83fc4) included "tags" file at the top level. Could you make an update to remove that from this PR, please?

.github/workflows/go.yaml
```diff
@@ -19,14 +19,28 @@ jobs:
         with:
           node-version: "10.x"
       - run: npm install
-  go-test:
+  lint:
```
dschott68, line 22: Since lint is only for ubuntu and not windows, did you want it in this PR or did you want to split it off into separate smaller PR?

codesearch/index/write.go
```diff
@@ -563,15 +563,16 @@ func (b *bufWriter) offset() uint32 {
 	return uint32(off)
 }
 
-func (b *bufWriter) flush() {
+func (b *bufWriter) flush() error {
```
dschott68, line 566: Should we care that the other callers of flush() besides Flush() are not handling the returned error?

flush_callers

dschott68 (approved):

PR #360, opened 2020-10-18
Pin Alpine version to 3.11.6

rfan-debug:
If you run the `docker build . -t nice/hound`, you will see the following information

```bash
OK: 12750 distinct packages available
  bzr (missing):
ERROR: unsatisfiable constraints:
    required by: world[bzr]
The command '/bin/sh -c apk update 	&& apk add go git subversion libc-dev mercurial bzr openssh 	&& go install github.com/hound-search/hound/cmds/houndd 	&& apk del go 	&& rm -f /var/cache/apk/* 	&& rm -rf /go/src /go/pkg' returned a non-zero code: 1
```

This is because `bzr` has become unmaintained due to its Python2 upstream: https://gitlab.alpinelinux.org/alpine/aports/-/tree/3.12-stable/unmaintained%2Fbzr

This PR pins the Alpine's version to 3.11.6 to avoid this issue. The latest is 3.12.

A minor improvement: 
I also removed the redundant COPY because it won't be used at all. When the user runs the docker command, it always redirects the `$(pwd)` in the host system to `/data` volumn for `houndd` command.

One more question for the future:
Should we keep supporting Bazaar? I heard that it is almost dead. https://www.reddit.com/r/Ubuntu/comments/2ywrif/has_bazaar_been_deprecated/

If we keep supporting Bazaar, we should consider [breezy](https://www.breezy-vcs.org/) as a successor of `bzr`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.10
+FROM alpine:3.11.6
 
 ENV GOPATH /go
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,8 @@
-FROM alpine
+FROM alpine:3.10
 
 ENV GOPATH /go
 
 COPY . /go/src/github.com/hound-search/hound
-
-COPY default-config.json /data/config.json
 
 RUN apk update \
 	&& apk add go git subversion libc-dev mercurial bzr openssh \


Dockerfile
```diff
@@ -1,11 +1,9 @@
-FROM alpine
+FROM alpine:3.10
```
ghmeier, line 1: What do you think about using 3.11.6? I think that's the last version of alpine prior to 3.12 at the moment.

rfan-debug: Yeah. I can switch to 3.11.6. It is kinda larger though.

rfan-debug: Done

Dockerfile
```diff
@@ -1,11 +1,9 @@
-FROM alpine
+FROM alpine:3.11.6
 
 ENV GOPATH /go
 
 COPY . /go/src/github.com/hound-search/hound
 
-COPY default-config.json /data/config.json
```
salemhilal, line 7: Yup, it would seem as though you are right about this line. If my understanding is correct (I'm a bit of a novice with Docker): 
1. We copy `default-config.json` to `/data/config.json`
2. We then mount `/data` as a volume, ignoring whatever was there anyways
3. We start hound expecting `config.json` to be in the `/data` directory

You don't know of a clever way to make the config default to `default-config.json` if there isn't a `config.json` in the mounted volume, do you?

salemhilal (approved):
I have one question, but I confirmed that the alpine upgrade works as expected locally.

Dockerfile
rfan-debug: If there is no `config.json` in that directory, the docker will quit with error because the entry point is like

```
ENTRYPOINT ["/go/bin/houndd", "-conf", "/data/config.json"]
```

You have to put match the name `config.json` in the `data` directory in the docker. When you run the docker, you mount the current path `$pwd` in your host system as the `data` volume for the docker. 

That's why we have to use `-v $(pwd):/data` in the docker run command.

rfan-debug:
> I have one question, but I confirmed that the alpine upgrade works as expected locally.

Actually that's a downgrade in the perspective of Dockerfile itself 😄 . With the current `Alpine` (no version argument), it will automatically pick the latest Alpine image, which does not contain `bzr` any more.

Dockerfile
salemhilal: Right, sorry — that all makes sense. I was wondering if there was an easy way to copy `data-config.json` over in such a way that it could be used if `config.json` was missing from `/data` (i.e. in the current working directory). I don't think that's a big deal though, since no one's running Hound just to use the default config.

salemhilal:
> Actually that's a downgrade in the perspective of Dockerfile itself
Yup! That makes total sense. That's what I get for reviewing PRs at 1am -_-

salemhilal:
Thank you for this contribution! We're trying to get regular docker deploys going again, and this will certainly help.

Dockerfile
rfan-debug: Yeah. Let's avoid using the default in the docker. It is harmful and hard for users to debug.

PR #362, opened 2020-10-25
Houndd version

### LINKED ISSUE #306: Is it possible to get houndd version ?
As **bin/go-bindata** provides version (bin/go-bindata -version), Is it possible to get bin/houndd version ?

/hound/bin# ./go-bindata -version
go-bindata 3.1.0 (Go runtime go1.5.3).
Copyright (c) 2010-2013, Jim Teeuwen.

kartikay101:
Fixes #306 
Display the version of houndd and exit

usage `houndd -version`

the version can be set while building 
`make VERSION=1.1.1` will set the version of houndd to 1.1.1

If no version is provided a default of 1.0.0 is used


salemhilal:
This is a good start! I think I'd prefer the version of the cli to be defined in the actual go runtime, rather than passed in as an environment variable at build-time. The version is more related to the contents of the code than when and how it's built. 

I think the approach that `gopass` uses makes a lot of sense. They expose a semver version struct as the source of truth: 
https://github.com/gopasspw/gopass/blob/6c94cbef056df696968a7cc4a8594b42c7318bb2/main.go#L253

It makes sense to me to expose the version in a similar way, either as a constant or through a getter function, in `hound/cmds/houndd/main.go`. Does that sound reasonable?

kartikay101:
That will be an easier thing to do, just setting 3 constants and returning a concat of those. The reason I chose build over that approach is because people tend to forget updating these and realize it a bit too late. Another thing that can be done in the current approach is to mandate version provision instead of defaulting it. 

Thoughts/ comments ?

kartikay101:
Made the relevant changes.
Used similar structure to gopass. It passes semver.Version instead of a string

Ways to provide version are:
1. Edit the main.go to change the version `Major=2 Minor=5 Patch=3` results in `2.5.3` as the version.
2. Provide it as VERSION flag in make. `make VERSION=2.4.3 results in `2.4.3` as the version (which overrides the version given inside the main.go file)

To see the version use `houndd -version`

Minor change, displayed `houndd v1.2.3` instead of just `houndd 1.2.3`

cmds/houndd/main.go
```diff
@@ -113,6 +116,18 @@ func runHttp(
 	return http.ListenAndServe(addr, m)
 }
 
+func getVersion() semver.Version {
+	ver, err := semver.Parse(version)
+	if err == nil {
+		return ver
+	}
+	return semver.Version{
```
salemhilal, line 124: nit: The latest release is 0.3.0, technically. Are you seeing v1.0.0 listed somewhere?

cmds/houndd/main.go
```diff
@@ -13,6 +14,7 @@ import (
 	"strings"
 	"syscall"
 
+	"github.com/blang/semver"
```
salemhilal, line 17: we've just added support for gomodules — are you ok with adding this to our `go.mod` file? I believe you will need to rebase master into this branch and force-push to see it.

salemhilal (changes requested):
Thank you for the updates! I left two blocking comments.

I think my idea about how releases would work are that they're intentionally done as part of a release. Can you tell me when we'd want to manually override just the release version, but not its code, during the build process? I'm not saying that there isn't a way, but I'm just having a hard time seeing it. The downside is that we might end up with a build that is labeled as one version, but has the source code for another.

Relatedly, I think some work for me to do here is to document our release process so that the steps for upgrading are clear.

At any rate, I'm gonna nudge some of the other maintainers about their thoughts on versioning. Aside from my comments and my thoughts on passing in a version flag, I like this PR!

kartikay101:
I agree with you, it isn't a good practice to have 2 sources of truth, thus the first approach ( only via build ) 

I agree with the idea to have it consistently available at both locations ( code + actual binary )

The only reason I went with the approach I took is to because the reference suggests so ( as they are using the version flag in  multiple locations ) 
Let me get this sorted and push it. 

P.S. Could you mark the pr as **hacktober-accepted** in the tags.

cmds/houndd/main.go
kartikay101: I will change it to 0.3.0

kartikay101:
Add `semvar to go.mod` 
Removed functionality  `add version through the makefile`

Only way to provide versioning is via code i.e Major Minor And Patch format.

kartikay101:
Could I nudge you to either **accept**/ **merge** this PR or just add a label as **hacktoberfest-accepted** as that would help me get my hacktober event completed.

 If there are any questions/ideas regarding this PR please feel free to reach out.

salemhilal:
@kartikay101 yes, totally. I planned on working through some of the backlog of issues today. I'll add the label for now, but I also hope to get this merged shortly as well. Thanks again for your contribution!

salemhilal:
After some local testing, this looks good to me! Thanks again, and I hope you snag that t-shirt! 
If you are looking for another issue before the deadline, there's a [short one](https://github.com/hound-search/hound/issues/363) that I just opened. You're welcome to it, but no pressure. I'll be working through some of the other pull requests before I try and implement that fix.

PR #369, opened 2020-12-03
Update README.md

aroach:
Add clarification of using file:// protocol for accessing locally pulled repos


salemhilal:
Thank you for updating! Much appreciated.

PR #371, opened 2021-01-22
Install dependencies in Dockerfile

yayachiken:
This fixes the Dockerfile such that building the container image is possible using something like:

    docker build .

Without adding the `semver` package, that build command would fail with the message

```
go/src/github.com/hound-search/hound/cmds/houndd/main.go:17:2: cannot find package "github.com/blang/semver" in any of:
	/usr/lib/go/src/github.com/blang/semver (from $GOROOT)
	/go/src/github.com/blang/semver (from $GOPATH)
The command '/bin/sh -c apk update 	&& apk add go git subversion libc-dev mercurial bzr openssh 	&& go install github.com/hound-search/hound/cmds/houndd 	&& apk del go 	&& rm -f /var/cache/apk/* && rm -rf /go/src /go/pkg' returned a non-zero code: 1
```

Note that I am not super-knowledgeable about the golang ecosystem. If it is possible to build the image without such changes, please tell me, I could then add the information to the README.

In addition, the pinned alpine image was bumped to the latest bugfix release, but the build works without that specific change.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.11.6
+FROM alpine:3.11.7
 
 ENV GOPATH /go
 
@@ -6,6 +6,8 @@ COPY . /go/src/github.com/hound-search/hound
 
 RUN apk update \
 	&& apk add go git subversion libc-dev mercurial bzr openssh \
+	&& cd /go/src/github.com/hound-search/hound \
+	&& go mod download \
 	&& go install github.com/hound-search/hound/cmds/houndd \
 	&& apk del go \
 	&& rm -f /var/cache/apk/* \


salemhilal:
Hey there! I'm sorry you're running in to issues here. We recently moved dependencies to go modules, and semver [is now listed as a dependency](https://github.com/hound-search/hound/blob/master/go.mod#L6) in our `go.mod` file. I'm relatively new to both go modules and Docker, though, so there may be an extra step we need to add to the dockerfile to ensure dependencies are installed. That being said, a `go get` step might not be the right approach here, since we'd need to add a `go get` for every other dependency.

I'll leave this PR open for now, but I may turn it into an issue if we can't figure an answer out shortly. Thank you for bringing this up!

yayachiken:
Thanks for the quick reply! Really appreciate it.

Yeah, I see your point. I searched the web a bit and it seems that individual calls to `go get` could be replaced by `go mod download`. I will probably test it next week (it's for work, and I cannot test on weekend for that reason)

If that doesn't work, IMHO it still would make sense to merge this PR, even if this is not how things are actually done in the golang best practives. As far as I can tell the Dockerfile is currently broken (even though I might be wrong, if there are some special parameters I can pass or something) and the fix would at least restore functionality to allow further refinement later on.

yayachiken:
OK, I tested the following two alternatives, and both create a working Docker image:

1. You can run `go mod vendor` in the hound repository before building the Docker container. In that case the `vendor` directory gets created with the dependencies specified in `go.mod`, they will get copied into the container, and `go install` will then use them. In that case, the `vendor` directory should probably be added to `.gitignore` and detailed build instructions containing the vendoring step be added to the README.

2. Before calling `go install` in the `Dockerfile` you can switch to directory `cd /go/src/github.com/hound-search/hound` within the container, and then execute `go mod download`. In that case the `Dockerfile` works without additional preperation steps.

I can modify the PR for either of the alternatives. @salemhilal which one would you think is the cleaner one?

yayachiken:
I also figured out why an old alpine base image was used. The `bzr` package seems to have been moved to a "community repo" in alpine Linux 3.12+.

As I am already working on the Dockerfile, I could try to fix that as well. Should it be part of this PR here or should I create a second one?

salemhilal:
@yayachiken Thank you for this investigation! I lean towards the 2nd approach since it encapsulates the setup in the Dockerfile, rather than in the documentation, but if you have strong feelings otherwise, I'd be open to the 1st approach. You're also welcome to swap out the base image if it seems straightforward to do — that would be very much appreciated.

yayachiken:
Thanks for the feedback! I have now implemented the 2nd approach. My branch is ready to merge.

Unfortunately I have now found out that `bzr` does not exist in the Alpine Linux repos anymore, at all. For that reason, the change is not at all straightforward. One would eventually have to either drop `bzr` VCS support in the Docker image OR switch to another base image which contains `bzr`, e.g. Debian. The latter alternative will probably not be too difficult to implement, but the Dockerfile would have to be rewritten to use `apt-get` instead of `apk` and stuff like that.

As I do not want to delay this PR, and don't know when I have the time, I think I will just open up an issue for now to track this.

salemhilal:
I've just confirmed that I can build and run that Dockerfile locally (I cleared my cache and such). This looks good to merge! Thank you again for reporting and fixing this problem.

PR #379, opened 2021-02-12
Adding a tip for windows users

StingyJack:
depending on where you put your hound data it could be competing with windows search indexer, which doesnt need to index this stuff. 

Also (for windows) its common for developers to have IDE and repo folders excluded form realtime virus scanning. That also competes with this indexer

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ Yup, that's it. You can proxy requests to the Go service through Apache/nginx/et
 
 ## Support
 
-Currently Hound is only tested on MacOS and CentOS, but it should work on any *nix system. Hound on Windows is not supported but we've heard it compiles and runs just fine, just remember to exclude your data folder from Windows Search Indexer and possibly the virus scanner.
+Currently Hound is only tested on MacOS and CentOS, but it should work on any *nix system. Hound on Windows is not supported but we've heard it compiles and runs just fine (although it helps to to exclude your data folder from Windows Search Indexer).
 
 Hound supports the following version control systems: 
 
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ Yup, that's it. You can proxy requests to the Go service through Apache/nginx/et
 
 ## Support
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -84,7 +84,7 @@ Yup, that's it. You can proxy requests to the Go service through Apache/nginx/et
 
 ## Support
 
-Currently Hound is only tested on MacOS and CentOS, but it should work on any *nix system. Hound on Windows is not supported but we've heard it compiles and runs just fine.
+Currently Hound is only tested on MacOS and CentOS, but it should work on any *nix system. Hound on Windows is not supported but we've heard it compiles and runs just fine, just remember to exclude your data folder from Windows Search Indexer and possibly the virus scanner.
```
salemhilal, line 87: Nit: I'm a little hesitant about telling people to have a virus scanner exclude the data folder, especially since it could be full of 3rd party code. I also added some minor phrasing tweaks, but I like this otherwise.
```suggestion
Currently Hound is only tested on MacOS and CentOS, but it should work on any *nix system. Hound on Windows is not supported but we've heard it compiles and runs just fine (although it helps to to exclude your data folder from Windows Search Indexer).
```

salemhilal (approved):
Thank you for this tip! I left one small tweak, and if that looks ok to you, I can merge it shortly.

README.md
StingyJack: I thought everyone excluded their IDE + Build tools, and source code folders.    The default windows antivirus/malware tool (MsMpEng.exe) was definitely battling hard to realtime scan the 1200+ repos that  20 searcher/indexers were downloading/processing.  The problem should be obvious enough to anyone else using this on windows if they look at the task manager so I have no complaints about the edit..

salemhilal:
@StingyJack With regards to the realtime scanner: I haven't been a windows dev in some time, so if you think that bit about antivirus exclusion rules would be valuable, we can totally put it back in. I'll merge this commit in for the time being.

PR #380, opened 2021-02-17
Add a warning for the default branch rename

salemhilal:
We're planning on renaming the default branch to `main`, and we want to give users a heads up a week in advance.


dschott68 (approved):
LGTM

PR #387, opened 2021-03-08
support github webhook

corbosman:
This patch adds support for github webhooks. This means that committing to a repository sends a webhook request to hound, which then re-indexes the repository. We've been using this in production for about 2 years on our Github Enterprise with 1000+ repos. Without this patch it was basically not possible as we had to re-index that many repos all the time. As per a previous issue I tried creating tests, but my go skills weren't enough. I couldn't get it to work, mainly because I think it means a bit of a rewrite of api.go.  

I'm sending this PR so someone can decide if it's useful at all, and if not, just close the PR. Since our company will shutter in a few months as least the code is out there.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -250,4 +250,51 @@ func Setup(m *http.ServeMux, idx map[string]*searcher.Searcher) {
 
 		writeResp(w, "ok")
 	})
+
+	m.HandleFunc("/api/v1/github-webhook", func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != "POST" {
+			writeError(w,
+				errors.New(http.StatusText(http.StatusMethodNotAllowed)),
+				http.StatusMethodNotAllowed)
+			return
+		}
+
+		type Webhook struct {
+			Repository struct {
+				Name string
+				Full_name string
+			}
+		}
+
+		var h Webhook
+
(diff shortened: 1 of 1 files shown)


salemhilal:
Hey @corbosman, I'm sorry to hear about your company, and I really appreciate you opening this PR. I or one of the other maintainers will try and take a look shortly. Seriously thank you, and best of luck.

corbosman:
Just for some additional information. We have this webhook configured on our Github Organisation level, so all repos automatically send a post to hound when things get updated. We don't actually index all repos, but that's ok, they just get a 404.  

* Payload Url:  https://hound.your.domain/api/v1/github-webhook
* Content-Type: application/json
* Secret:  <empty>.  (could be a future enhancement to allow for a secret, we didn't need it internally)

There's 2 important fields in the github post, name and full_name. Full_name includes the organisation. It may need some tweaks if you're using github.com, im not sure. Should all be relatively easy to modify. 

The main thought behind it is to only index when a repo changes, and then index on-the-fly. 

Sorry I couldn't provide a better patch, Go is not one of my languages, but since I maintain the docker swarms where our hound runs, we had to do something about the large amount of indexing. (it's close to 1500 repos).

Issue #368, opened 2020-11-03
Latin charset=ISO-8859-1 parse


ViljarVoidula:
Hi there!

We have localized git, which store translations to languages , therefore ÖÄÕÜ etc ISO-8859-1 compliant files seem to be excluded as "Invalid UTF-8". How to overcome this problem, unfortunately i have no experience with GO to help with PR right now. Help is appreciated.

salemhilal:
Hey there! I would not be surprised to hear that Hound has problems with text encoding. I'm still a bit newer to the codebase myself. Can you share the exact error you are seeing, including any stack trace, and the steps you are taking to far it to appear? Also, does Hound crash outright or does it ignore that particular repository?

ViljarVoidula:
@salemhilal - Nothing crashes, so its rather positive problem , i can just see these files @ excluded_files.html in client and unable to search for contents. I can collect logs maybe in weekend

Issue #243, opened 2017-03-01
Encoding issues with ANSI files.


AlexFielder:
Hi all,

We're now using Hound to index a fairly large codebase >1.5GB and I just noticed that there are 1,000s of files being ignored if they have ANSI encoding instead of UTF-8. They appear in the excluded_files.json file

Is this something that can be fixed within Hound or will I likely have to figure out how to change the encoding of all these files?

Thanks,

Alex.

AlexFielder:
For anyone else having the same issue, here is the (quickest) fix I could come up with:

1. Install Notepad++ (NP++) 32 bit (important because there's currently no plugin manager for the default x64 installation from chocolatey.org!)
2. Install the Python Script plugin for NP++
3. Create new script: ![2017-03-22 11_08_11-c__projects_go_hound no cm_error log - notepad administrator](https://cloud.githubusercontent.com/assets/167413/24194986/ecca0534-0eef-11e7-9382-c3d444bff400.png)
4. Inside of the new script add the following:
```python
import os;
import sys;
filePathSrc="C:\\CM"
for root, dirs, files in os.walk(filePathSrc):
    for fn in files:
      if fn[-3:] == '.cm' :
        notepad.open(root + "\\" + fn)
        console.write(root + "\\" + fn + "\r\n")
        notepad.runMenuCommand("Encoding", "Convert to UTF-8")
        notepad.save()
        notepad.close()
```
(You can set up the script to only work on specific file extensions like I have or change it to suit.)
5. Save and close the script file, then NP++ itself. (This is necessary to get Python Script to pick up the newly saved script file)
6. Reopen NP++ and select your script file (whatever it's called):
![2017-03-22 11_11_56-c__users_alex fielder_dropbox_scripts_convert_to_utf-8 py - notepad administr](https://cloud.githubusercontent.com/assets/167413/24195118/70cfb0e0-0ef0-11e7-9279-1a9d82bc7eec.png)
7. Go and grab a coffee, because if (as I did) you have 20,000+ files to check it will take some time. Whilst the script is running the system will fight with NP++ for focus as it cycles through the folder structure you pointed it at.

FWIW: I am using an NVMe Samsung SSD and this script barely taxes it; if you have a regular spinny HDD then you might need to be selective where you point it.

Issue #78, opened 2015-02-24
Issue with UTF16 files


mattdurham:
We have a mixture of UTF8 and UTF16 files. Hound seems to search them correctly, as in it will say 16 files searched, but only show 4. When I go and manually switch save as UTF8 and check in then the files will show in the UI.

jklein:
Interesting, thanks for the report. I haven't had a chance to reproduce yet, but it seems like this should be easy to confirm and potentially easy to fix.

kellegous:
I think we can just sniff for a BOM in the first two bytes and re-encode the file to utf-8 on the fly. It will add some overhead to the indexing for repos with utf-16 but that's unavoidable anyway because Go's strings are utf-8 internally.

PR #388, opened 2021-03-11
Add fallback-encoding per-repo option for non-utf8 text files

tgulacsi:
This allows to specify an alternate charset (fallback-encoding) per repository.


tgulacsi:
May help #78, #243 and #368, and my ISO8859-2 case :)

salemhilal (commented):
Hey, thank you for contributing! Encoding stuff can be pretty scary — can you add some tests to your changes? We're putting some eyes on this PR now otherwise, but we'd want some test coverage before we merge.

index/index.go
```diff
@@ -14,13 +14,14 @@ import (
...
 	"github.com/hound-search/hound/codesearch/regexp"
+	"golang.org/x/text/encoding"
 )
 
 const (
 	matchLimit               = 5000
 	manifestFilename         = "metadata.gob"
 	excludedFileJsonFilename = "excluded_files.json"
-	filePeekSize             = 2048
+	filePeekSize             = 1 << 20
```
jacobrose, line 24: Hello! Why this big jump in filePeekSize?

tgulacsi: That pesky comment that contains á that breaks UTF-8 detection may be at the end of the file.
This is not a full solution, just a bet that source files should be less than a megabyte long.

A full solution could be retrying the whole indexing on decoding error.
But that felt more involved - though I could try it if you confirm that'd be better.

salemhilal:
Hey! We're meeting now and generally like this PR. We have two asks before we merge it: 

1. Can you add some tests? 
2. Increasing `filePeekSize` might impact startup time. Can you give us any details on how much of an impact you are seeing, and if possible, a rough estimate of the size of your codebase?

Thank you again!

tgulacsi:
Sorry for my slow reply, but now I've added a test case, and implemented the more robust retrying logic.
PTAL.

tgulacsi:
May I assist reviewing this PR?

Issue #381, opened 2021-02-19
Use //go:embed from Go 1.16


simon04:
https://golang.org/pkg/embed/

>  Package embed provides access to files embedded in the running Go program.
>
> Go source files that import "embed" can use the //go:embed directive to initialize a variable of type string, []byte, or FS with the contents of files read from the package directory or subdirectories at compile time. 

https://golang.org/doc/go1.16#library-embed

salemhilal:
Hey! Just to clarify, are you proposing that this replaces `ui/bindata.go` and the `go-bindata`package?

simon04:
Exactly, yes.

salemhilal:
Ok, cool. We aren't going to prioritize this for the time being, but we are open to a PR.

moson-mo:
Seems like the main blocker is that with the switch to `embed`, a simple `go install ...` won't work anymore / people would need to have `npm` installed as well when compiling hound?

I'd agree with 3 points mentioned here: https://github.com/hound-search/hound/pull/430#issuecomment-1172640962

Regarding _Commit the ui build artifacts to the repo_:  
That's actually what's already happening with `bindata.go`. Switching to `embed` would just shift that to different files.
(which I find somewhat dangerous.  It's much easier to hide malicious stuff in "binified" go files or minified JS) 😉

dschott68:
resolved by https://github.com/hound-search/hound/pull/460 and thank you!

PR #392, opened 2021-03-30
Replace go-bindata with //go:embed from Go 1.16

simon04:
Fixes https://github.com/hound-search/hound/issues/381.

cc: @salemhilal


Makefile
```diff
@@ -9,24 +9,21 @@ endif
...
-.build/bin/go-bindata:
-	GOPATH=`pwd`/.build go get github.com/go-bindata/go-bindata/...
-
-ui/bindata.go: .build/bin/go-bindata node_modules $(wildcard ui/assets/**/*)
-	rsync -r ui/assets/* .build/ui
+ui/.build/ui: node_modules $(wildcard ui/assets/**/*)
+	mkdir -p ui/.build
+	rsync -r ui/assets/* ui/.build/ui
 	npx webpack $(WEBPACK_ARGS)
-	$< -o $@ -pkg ui -prefix .build/ui -nomemcopy .build/ui/...
```
simon04, line 29: Maybe this rule needs some tuning for proper rebuilds.

salemhilal: Can you elaborate more on what you  mean here?

.github/workflows/go.yaml
```diff
@@ -1,10 +1,10 @@
-on: [push]
+on: [push, pull_request]
 jobs:
   go-build:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: ["1.14", "1.13", "1.12"]
+        go: ["1.16"]
```
salemhilal, line 7: While it's true that we probably don't need to be supporting Go 1.12 or 1.13, I'm wary of dropping support for everything but the latest version. I believe the two most recent Go versions are officially supported. This might make a little more sense once Go 1.17 is out.

Makefile
```diff
@@ -9,24 +9,21 @@ endif
...
 
-.build/bin/go-bindata:
-	GOPATH=`pwd`/.build go get github.com/go-bindata/go-bindata/...
-
-ui/bindata.go: .build/bin/go-bindata node_modules $(wildcard ui/assets/**/*)
-	rsync -r ui/assets/* .build/ui
+ui/.build/ui: node_modules $(wildcard ui/assets/**/*)
+	mkdir -p ui/.build
+	rsync -r ui/assets/* ui/.build/ui
 	npx webpack $(WEBPACK_ARGS)
```
salemhilal, line 26: I think the original motivation around using bindata in the first place is so that we could ensure built static assets were present and usable immediately after `go install`, without requiring an additional build step / installing `node_modules` / etc. I'm admittedly fuzzy about whether or not Go Modules have a post-build lifecycle step, but I think that context is important for this PR.

Makefile
simon04: I'll try. Since `ui/.build/ui` is a directory, I'm not sure under which circumstances make will figure out that the target needs to be rebuilt. Maybe it would be better to only check the JavaScript bundle generated by webpack?

Makefile
jrarmstro: I have had success with [pkger](https://github.com/markbates/pkger).

.github/workflows/go.yaml
prochac: Walking around and reminding that Go 1.17 was released.

dschott68:
replaced by https://github.com/hound-search/hound/pull/430

Issue #375, opened 2021-01-25
Docker Image: Handle the bzr package being removed from Alpine Linux


yayachiken:
As I found out on the discussion PR #371, Alpine Linux dropped the `bzr` package in the releases 3.12 and onwards. For that reason, the Dockerfile currently pins the latest Alpine Linux 3.11.x release.

Alpine Linux 3.11.x is still supported, but when that gets EOL this would become a problem, as the Docker image could not be built anymore with a current image receiving security fixes etc.

I see the following two solutions to the problem:

1. Remove `bzr` support from the Docker image by simply omitting that package from the Dockerfile. This would obviously be a loss of functionality and is probably not the way to go unless no other alternatives are possible.
2. Base the Dockerfile on another base image which still contains `bzr`. In that case the Dockerfile would have to be rewritten to replace distribution specific commands like `apk` with the equivalents of the newly chosen base image distribution.

salemhilal:
(This could just as well be marked as a bug, since this would be breaking after Alpine Linux 3.11.x is EOL'd)

foragerr:
Do you folks have a preferred base image as an alternative to alpine?

I have an unrelated DNS resolution issue with running the alpine based image in AWS Fargate. I'm about to create my own hound image from scratch, happy to go with a base image you may have an affinity for, and contribute it back.

foragerr:
aside: The comment on this alpine commit says to use breezy as a replacement for bzr

>Python2 only with no upstream. Most distros replaced it with breezy
which uses python3 and if there is demand for bzr then breezy and its
dependencies should be packaged instead

https://gitlab.alpinelinux.org/alpine/aports/-/tree/3.12-stable/unmaintained%2Fbzr

dschott68:
resolved with https://github.com/hound-search/hound/pull/429

PR #393, opened 2021-04-02
replace alpine with debian based golang base image

foragerr:
attempt at fixing #375 
work in progress

I don't like that the image ends up much larger than the original alpine based one

```
etsy/hound   latest   66ef348e6d82   2 years ago     138MB
hound        temp     dfa1680178e9   9 hours ago     955MB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,20 @@
-FROM alpine:3.11.7
+FROM golang:1.16-buster
 
 ENV GOPATH /go
 
 COPY . /go/src/github.com/hound-search/hound
 
-RUN apk update \
-	&& apk add go git subversion libc-dev mercurial bzr openssh tini \
+RUN apt update \
+	&& apt upgrade -y \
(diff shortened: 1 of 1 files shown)


foragerr:
Hmm the go.yaml workflow does not exercise the dockerfile. Are there checks I should run against dockerfile changes?

nickmoorman:
Hi!  Sorry for the delayed response, and thanks for your patience.  Switching distros seems a little scary, and it also looks like it results in a pretty non-trivial size increase in the resulting Docker image.  Looking at the issue you linked to, it seems like the bigger problem is to make sure Bazaar repos can still be supported as OS version upgrades are applied to Hound.  Instead of switching over to Debian, would you mind looking into what it would take to replace `bzr` with `breezy` and update Hound's [VCS adaptor](https://github.com/hound-search/hound/blob/b1bc7538503286dbf3e9dd7b98c557060c327884/vcs/bzr.go) accordingly?

dschott68:
replaced by https://github.com/hound-search/hound/pull/429

PR #401, opened 2021-06-01
Automatically build docker image and publish to ghcr.io

inful:
This should make publishing docker images a lot simpler. You need to enable the "Improved container support" feature preview, though. It should push images to https://ghcr.io/hound-search/hound:[TAG]


salemhilal:
We're wrapping our heads around this, but it seems pretty useful! Thank you for the contribution.

inful:
> We're wrapping our heads around this, but it seems pretty useful! Thank you for the contribution.

NP - You can examine the resulting image available from my fork at ghcr.io/inful/hound:latest
I have been running this without any issues.

PR #402, opened 2021-07-01
Now support https://user:pass@server:port/... URLs

ChloeTigre:
When generating the URLs, strip the security credentials.

This allows to pass a config such as (only the relevant snippet)

```json
"matchalunatic / hound-private" {
      "url": "https://secretuser:secretpass@gitlab.com/matchalunatic/hound-secret.git",
      "url-pattern": {
        "anchor": "#L{line}",
        "base-url": "{url}/blob/master/{path}{anchor}"
      }
}
```

and to strip the secretuser:secretpart part of the URL when browsing away from hound, so that people use their own credentials when browsing.


salemhilal:
Hey! I don't likely think this is a feature we would want to support. It's a lot safer to give the machine that's running Hound [SSH permissions](https://docs.gitlab.com/ee/ssh/) to your private repos, rather than putting a username and password in a config (we use Hound with private repos ourselves). Let me know if this helps, and if there's some use case we aren't thinking of.

ChloeTigre:
Hello @salemhilal 

Sure, I can explain! We are an org with 100+ engineers and we run on GitLab. Our Hound setup is based on k8s, scales super well and autogenerates config (I'd be glad to contribute the config maker script as well).

As you may know, GitLab supports SSH keys only from users and not for project or group-related machine credentials. Therefore, we (and other companies) have a situation where 1/ we want to access private repositories but 2/ do not want to grant access based on one specific set of user credentials, because employees as you know well come and go.

This is a go-to situation for Gitlab group tokens and this is why we use them. But we don't want them to leak when the users make use of Hound, because they should be able to access GitLab with their own credentials.

I would like to merge this because this is a simple enough way of achieving our goals and it does not involve managing a separate config for credentials, but if you really can't accept that then I will open another merge request that does the following:

- keep the config as-is
- add another config file with a mapping between a repository and file-based credentials (to store the passwords separately)
- pass the credentials to the HTTP requester

but really this solution in the current MR is better.

Right now Hound has no way that I know of to checkout private Gitlab repositories over HTTPS without leaking the credentials, and so it will be (on GItlab) necessarily bound to a human user and not to an organizational machine user.

Hence this MR and the reason why, IMO, you should reconsider.

Thanks :)

salemhilal:
Hey @ChloeTigre, thanks for taking the time to put this response together!

> As you may know, GitLab supports SSH keys only from users and not for project or group-related machine credentials.

It _looks_ like public deploy keys [might be SSH keys for machines](https://docs.gitlab.com/ee/user/project/deploy_keys/#public-deploy-keys). The use case they mention is for CI/CD instances, which functionally fit the same sort of profile as a Hound server would. Is this something that would solve your problem?

Our concern is less about HTTP authentication and more about Hound having to read and manage secrets. If the repo URL becomes a place where secrets can commonly get leaked, it could lead to security holes in future releases.

PR #403, opened 2021-07-05
Added authenticated-url support in config

ChloeTigre:
Rationale: need to be able to use HTTPS + Access Tokens for some private GitLab repositories, but also need not to share a secret with the Hound UI users.

Implementation: a new optional key  "authenticated-url" sitting besides "url". This is not transmitted to the client.

Tests: added a config block using it but no specific test.

Tech:
Added a SecretString type just for this because SecretMessage was quirky to use with strings

More context: purpose replaces https://github.com/hound-search/hound/pull/402

### DIFF
--- a/config-example.json
+++ b/config-example.json
@@ -1,92 +1,103 @@
 {
-    "max-concurrent-indexers" : 2,
-    "dbpath" : "data",
-    "title" : "Hound",
-    "health-check-uri" : "/healthz",
-    "vcs-config" : {
+    "max-concurrent-indexers": 2,
+    "dbpath": "data",
+    "title": "Hound",
+    "health-check-uri": "/healthz",
+    "vcs-config": {
         "git": {
-            "detect-ref" : true
+            "detect-ref": true
         }
     },
-    "repos" : {
-        "SomeGitRepo" : {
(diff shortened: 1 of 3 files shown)


salemhilal:
Hey @ChloeTigre, I'm going to close this PR as well for the same reasons mentioned in the previous PR: 
https://github.com/hound-search/hound/pull/402#issuecomment-874800679

In short, it looks like what you're describing is possible with existing Gitlab access keys, and I'd still rather not be in the business of managing secrets. I'm open to talking about this more if you don't think SSH machine keys will solve the problems you're running into.

ChloeTigre:
Hi,

SSH machine keys can never solve that when you are deploying to serverless environments/k8s clusters and the likes.

SSH machine keys are bound to machine accounts, which are debatably not a great practice. GitLab supports Access tokens for reasons like this.

In this PR, I give provisions for securely holding these secrets while not exposing them. They are never exposed to the user but are used to fetch stuff still. This gives search access to private repositories without machine keys. 

Machine keys are also an annoyance as you cannot have several in order to access different organizations. I cannot have 2 different machine keys to access 2 separate projects on gitlab.com without hacks, while I can have different access tokens for different projects securely and without hacks.

This is why you should merge this :)

PR #404, opened 2021-07-27
Give file links a target of blank

hoorayimhelping:
Individual links to line numbers open in a new target. Links to the file (at the top of the list of line number links) open in the same window / tab. This PR makes them behave consistently with line links.

Add rel="noopener noreferrer" to _blank links


salemhilal (approved):
Thank you for this!

PR #405, opened 2021-08-21
Update README to use ghcr.io

davegallant:
Currently, the README suggests running an outdated version of hound. There's been a lot of good work done in the past [3 years](https://hub.docker.com/r/etsy/hound/tags).

The container registry was updated in https://github.com/hound-search/hound/pull/401

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,15 +56,15 @@ A complete list of available config options can be found [here](docs/config-opti
 
 ### Using Docker (1.4+)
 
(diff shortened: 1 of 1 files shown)


dereckson:
Probably a good idea to update it, as it can be confusing to use etsy/hound when it creates issues like described in #407.

salemhilal:
I have some time carved out on Wednesday. I'll try to publish a new image and update the docs. Thanks for your patience on this.

salemhilal:
@dschott68 and I went over these docs and they seem to work as expected, at least on my laptop (macOS 12.0.1, Apple M1 Pro). Thanks again @davegallant.

PR #408, opened 2021-09-16
Add support for manually setting a limit on results

### LINKED ISSUE #217: Exceed limit results
Hi guys, as I can see here https://github.com/etsy/hound/blob/203bd0c7601e28de8c4ae4e922bc0b76938af268/index/index.go#L198 - results are just dropped if they exceed limit (which is constant - 5000).

It is possible to return what was collected so far, instead of just dropping everything. Or, alternatively you can let user set max number of results he wants, of course limited by server's limit too.

For me both are better than what we have now, what do you think?

AlexTalks:
* Added a query parameter to allow user to specify max results, rather
  than a hard limit at 5000, for each index.
* Added support for a default max result limit in the config.  If unset,
  it remains at the former default of 5000 results per index.
* Fixed issue where no results would be returned if the limit was
  exceeded - now it simply returns the results up to the limit.

Resolves #217

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
-FROM alpine:3.11.7
+FROM alpine:3.16
 
 ENV GOPATH /go
 
 COPY . /go/src/github.com/hound-search/hound
 
 RUN apk update \
-	&& apk add go git subversion libc-dev mercurial bzr openssh tini build-base\
+	&& apk add go git subversion libc-dev mercurial breezy openssh tini build-base \
 	&& cd /go/src/github.com/hound-search/hound \
 	&& go mod download \
 	&& go install github.com/hound-search/hound/cmds/houndd \


api/api.go
```diff
@@ -128,11 +129,25 @@ func parseAsUintValue(sv string, min, max, def uint) uint {
 		return max
 	}
 	if min != 0 && uint(iv) < min {
-		return max
+		return min
```
salemhilal, line 132: Nice catch 🔥

api/api.go
```diff
@@ -18,6 +18,7 @@ import (
 const (
 	defaultLinesOfContext uint = 2
 	maxLinesOfContext     uint = 20
+	maxLimit              int  = 100000
```
salemhilal, line 21: Out of curiosity, is this number informed by anything or is it just reasonably large?

index/index_test.go
```diff
@@ -59,6 +61,41 @@ func TestSearch(t *testing.T) {
 	}
 }
 
+func TestSearchWithLimits(t *testing.T) {
```
salemhilal, line 64: Thank you a million times over for adding tests for this.

api/api.go
```diff
@@ -128,11 +129,25 @@ func parseAsUintValue(sv string, min, max, def uint) uint {
 		return max
 	}
 	if min != 0 && uint(iv) < min {
-		return max
+		return min
 	}
 	return uint(iv)
 }
 
+func parseAsIntValue(sv string, min, max, def int) int {
```
salemhilal, line 137: Could I trouble you to write a few tests on the bounds checking logic for this and for `parseAsUintValue`?

salemhilal (commented):
Thank you for this PR! I especially appreciate that you've added a configuration option for this setting. 

I left a few minor comments. I also want to get some feedback from the rest of the team about whether or not we should allow for a limit, but still warn once that limit has been reached. If you can handle the few minor comments in the mean time, this PR should be good to merge.

PR #235, opened 2017-01-31
ui: collapse and expand files

tanyunshi:
Refers to https://github.com/etsy/hound/issues/182, https://github.com/etsy/hound/pull/235 and https://github.com/etsy/hound/issues/201: New feature for search quick navigation.

 - Click on the button [expand all] or [collapse all] to expand/collapse all
 - Click on a repository title to expand/collapse all the files of the repo
 - Click on a file title to expand/collapse the file content

demo

PR #417, opened 2022-03-21
ui: collapse and expand files

vracini:
Hi, this contains the changes from #235 brought up to date.


ui/assets/js/common.test.js
```diff
@@ -1,4 +1,4 @@
-import { EscapeRegExp, ExpandVars, UrlToRepo } from "./common";
+import { EscapeRegExp, ExpandVars, UrlToRepo, UrlParts } from "./common";
```
salemhilal, line 1: thank you! I'm surprised tests were passing before this import was added 0_o

salemhilal (commented):
I'm into this change. I'm going to run it by another maintainer before we merge, but this is definitely a much-appreciated contribution.

ui/assets/js/common.test.js
vracini: Same here, looks like npm tests are not set-up in the [workflow](https://github.com/hound-search/hound/blob/main/.github/workflows/go.yaml#L81)

PR #421, opened 2022-05-23
Attempt auto-generated files being hidden by default

ivantsepp:
This PR introduces a new feature to auto hide autogenerated files. It builds upon the recent commit of being able to collapse files (https://github.com/hound-search/hound/commit/c1ffa46a65030bd59edf1584586a849a08dc5b6b). The idea is to parse out autogenerated file patterns from `.gitattributes` ([link](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github)) and automatically collapse matching files in the UI. 
Hound

**Note**
This is a PR just to spark discussion! I'm not sure what the best approach is as there are many options. We can ask questions like:

- automatically exclude these files in `excluded_files.json` (should speed up searches as these files aren't indexed)
- add a button to hide and reveal autogenerated files
- allow this list to be expressed in configuration

Thoughts on a feature like this? And what would the best approach be? Thanks!


salemhilal:
I love this idea! Here are some initial thoughts:
- I really like the idea of parsing out `.gitattributes`. `linguist-generated` annotations are already used to do essentially the same thing in Github, and mirroring that behavior here seems intuitive.
- I am not sure if it makes sense to put this behavior behind a config param or not. I'm erring towards no, at least unless it becomes an issue for someone.
- I think it might be a good idea to still index these files, but leave them collapsed (similar behavior to Github). Having files be excluded entirely feels to me like something you'd want to opt into separately. I don't usually want to see the output of generated files, but I do on occasion.

That's all my first pass on the details of implementation. Generally though, this feels like a feature I didn't realize I needed. I'm gonna read through the PR now.

searcher/searcher.go
```diff
@@ -407,16 +407,18 @@ func newSearcher(
...
 	rev, err := wd.PullOrClone(vcsDir, repo.Url)
 	if err != nil {
 		return nil, err
 	}
 
+	opt := &index.IndexOptions{
+		ExcludeDotFiles:           repo.ExcludeDotFiles,
+		SpecialFiles:              wd.SpecialFiles(),
+		AutoGeneratedFilePatterns: wd.AutoGeneratedFilePatterns(vcsDir),
+	}
```
salemhilal, line 420: Just to clarify, you swapped the order of this block so that we don't create `opt` if there's an error pulling?

vcs/git.go
```diff
@@ -151,6 +154,29 @@ func (g *GitDriver) SpecialFiles() []string {
...
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		matches := autoGeneratedFileRegexp.FindStringSubmatch(scanner.Text())
+		if len(matches) == 2 {
+			pattern := strings.ReplaceAll(matches[1], "**", "*")
+			pattern = strings.ReplaceAll(pattern, "*", ".*")
+			filePatterns = append(filePatterns, pattern)
```
salemhilal, line 173: Nit: Could you add a test for this pattern-to-regex conversion?

ui/assets/js/hound.js
```diff
@@ -693,7 +693,7 @@ var ContentFor = function(line, regexp) {
 
 var FileContentView = React.createClass({
   getInitialState: function() {
-    return { open: true };
+    return { open: !this.props.isAutoGenerated };
```
salemhilal, line 696: It could be nice to put an `[AUTOGENERATED]` badge next to the file path. That's  not a blocker of course, but it would be a nice visual indication about why the file is collapsed.

ivantsepp: Great idea, will do!

ivantsepp:
Thanks for taking a look and glad you think this feature is worthy of being added! I agree we can follow Github's approach where we hide but still index/search on these autogenerated files. 

I'll look into adding a badge, adding documentation into the README, and adding config override! 

Actually for the glob-to-regex logic, I'm not sure it's the right approach anymore. I just realized that you could nest `.gitattributes` in your directories so parsing out only one file that's in the root isn't full coverage. What about running a shell command like `git ls-files | xargs git check-attr linguist-generated`. Then we would get specific path names and we wouldn't need to regex anymore (we could just do a simple string comparison). Thoughts on this approach? I like this too since we can use `AutoGeneratedFiles` instead of `AutoGeneratedFilePatterns` (shorter name). 

(sorry for not responding earlier!)

searcher/searcher.go
ivantsepp: ah i swapped it because i needed `wd.PullOrClone` to be executed first so that when `wd.AutoGeneratedFilePatterns(vcsDir)` is called, it can find the `.gitattributes` file. `wd.PullOrClone` will populate the directory so without it, the first run of this flow will not see accurate results for autogenerated files

salemhilal: ahh gotcha, that makes sense.

ivantsepp:
Flushed out that approach and added a badge!

Hound


Added config override and added docs for that. Though I wasn't sure if I should add notes in the main README about this?

vcs/git.go
```diff
@@ -151,6 +152,51 @@ func (g *GitDriver) SpecialFiles() []string {
...
+	// Split by NUL and we expect the format: <path> NUL <attribute> NUL <info> NUL
+	tokens := bytes.Split(out, []byte{0})
+
+	for i := 2; i < len(tokens); i+=3 {
+		if string(tokens[i]) == "true" && string(tokens[i-1]) == autoGeneratedAttribute {
+			files = append(files, string(tokens[i-2]))
+		}
+	}
+
+	if err := filesCmd.Wait(); err != nil {
```
ivantsepp, line 192: Followed the pattern in the example in https://pkg.go.dev/os/exec#Cmd.StdoutPipe so hopefully this is right

ivantsepp:
friendly bump! was hounding the other day and was reminded that this feature would be cool!

salemhilal:
@ivantsepp Thank you for the bump! I think these updates look good. I'll float this PR around with the rest of the folks to see if there are any objections, but otherwise I'll merge this and cut a release. Thank you again!

dschott68 (approved):

PR #461, opened 2023-05-06
Add a `local` vcs driver

mrexodia:
Closes #344.

I tested it locally with the leaked XP source code (took maybe 6 minutes of indexing) and another smaller repo.

The `url` with `file://` is bothering me a little bit, it could just be the absolute path but this works.

TODO: a few comments from https://github.com/hound-search/hound/pull/344#pullrequestreview-520875229 still apply:

- [ ] Validate that the driver's initialized with a url that begins with file://, and return an error from newFake() otherwise. <- I did this check somewhere else, will move it.
- [ ] Add a test file. We don't have any strong integration test patterns yet, but at least validating ^ logic would be a start. It would be extra nice to test the hashing logic as well. I'm also very open to integration test thoughts, if you have any. <- I can add an empty file, but there is no infrastructure in place for these tests. Doing this properly would be a major undertaking for testing 100 lines of (mostly) trivial code, which seems silly.
- [ ] Add doc blocks

I removed the dependency on the `tar` command (which doesn't exist on Windows) and instead replaced it with the code used by `go mod`. The `watch-changes` option is set to `false`, in which case it will use the hash of the URL. Hashing all the files per default seems wasteful, since it's unlikely you are indexing code that frequently changes, but is not checked in.


mrexodia:
Ping

vcs/local.go
```diff
@@ -0,0 +1,105 @@
...
+		if err := json.Unmarshal(b, &d); err != nil {
+			return nil, err
+		}
+	}
+
+	return &d, nil
+}
+
+// Adapted from: https://github.com/golang/mod/blob/ad6fd61f94f8fdf6926f5dee6e45bdd13add2f9f/sumdb/dirhash/hash.go#L44
+func myHash(files []string, open func(string) (io.ReadCloser, error)) (string, error) {
```
salemhilal, line 39: It looks like this function is identical to the `Hash1` function exposed in dirHash, except yours usds sha1. Is there a reason you didn't go with `Hash1` here? (If there is, I'm probably going to ask that you add it in comment here)

salemhilal (commented):
Hey! Thanks for being patient with me on this review. I left a few comments. My evergreen one is for there to be a few comments in the parts where the rationale may be hard to infer by newcomers, but the one I'm particularly curious about is your implementation of `myHash` in place of the `Hash1` function. Let me know what your thoughts are there, but otherwise this PR is looking good to me.

vcs/local.go
mrexodia: I was doing it with some friends on stream. If I recall correctly I didn't use `Hash1` verbatim because it's a base64 string and this doesn't play well with the URLs. The reason for using sha1 was that (supposedly) this is the fastest hash algorithm as implemented in go. Will add some comment!

index/index.go
```diff
@@ -375,7 +375,12 @@ func indexAllFiles(opt *IndexOptions, dst, src string) error {
 	}
 	defer fileHandle.Close()
 
-	if err := filepath.Walk(src, func(path string, info os.FileInfo, err error) error { //nolint
+	if fi, err := os.Stat(src); err == nil && fi.Mode()|os.ModeSymlink != 0 {
+		if s, err := os.Readlink(src); err == nil {
+			src = s
+		}
+	}
```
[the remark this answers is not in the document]
mrexodia: Yes, it symlinks the local folder instead of copying it so it has to follow them. Will add a comment!

mrexodia:
Alright, I added the comments you requested let me know if there is anything else!

salemhilal (approved):
Thank you for the updates! This looks really good to me. I appreciate your patience with all the back-and-forth.

salemhilal:
Ah, one update to ask of you @mrexodia: it looks like there's a linting error.
https://github.com/hound-search/hound/actions/runs/5215441474/jobs/9429045032?pr=461

mrexodia:
> Ah, one update to ask of you @mrexodia: it looks like there's a linting error. https://github.com/hound-search/hound/actions/runs/5215441474/jobs/9429045032?pr=461

I suppressed the error as was done before. Just FYI this linter is extremely annoying. The job message was unhelpful in explaining how to reproduce the error and it took me 30 minutes to find a command that worked:

```
golangci-lint -v run --disable-all -E staticcheck --timeout 1h
```

Ideally the command to reproduce (just) the errors would be in the logs of the `lint` job, including installation instructions.

salemhilal:
Thank you for these changes! I'll make a note to see what we can do about improving the linting process. Sorry about the frustration there.

PR #478, opened 2023-11-23
Example of manual modifications required to use an ssh key as input

1Mark:
Documentation is lacking for how to use ssh key as input. It would be nice to explain how people should use ssh keys with this project.

As a minimum I needed to update the Dockerfile and add an entrypoint.sh to integrate with tini. Not sure if any existing documentation suggests how people should use this.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -8,7 +8,7 @@ import (
 )
 
 const (
-	defaultMsBetweenPoll         = 30000
+	defaultMsBetweenPoll         = 300000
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 const (
-	defaultMsBetweenPoll         = 30000
+	defaultMsBetweenPoll         = 300000
```
1Mark, line 11: It would be nice to provide this at a global level as well as a per repo config.

